PR #42, opened 2024-01-24
Small tweaks to make nanotron a bit more flexible

thomwolf:
be able to intialize torch distributed outside of the Parallel Context + small config tweaks


NouamaneTazi (commented):
LGTM!

PR #56, opened 2024-02-05
[FP8 Training] A single forward and backward pass for a linear in FP8

xrsrke:
This PR contains a single FP8 Linear layer supporting both forward pass and backward pass. The next PRs will enable FP8 end-to-end training.

image


src/nanotron/fp8/constants.py
```diff
@@ -0,0 +1,34 @@
+from enum import Enum, auto
+
+import torch
+
+FP8_GPU_NAMES = ["h100", "rtx 4090"]
+
+# FP8_DTYPES = [torch.fp8e4m3, torch.fp8e5m2]
```
NouamaneTazi, line 7: Are dtypes such as 'torch.fp8e4m3' deprecated?

src/nanotron/fp8/constants.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+DEFAULT_SCALE = 0.1
+DEFAULT_FP8_FORMAT = torch.int8
+
+
+class DTypes(Enum):
+    FP8E4M3 = auto()
+    FP8E5M2 = auto()
+    kfloat16 = auto()
```
NouamaneTazi, line 23: i would expect this to be uppercase as well?

src/nanotron/fp8/constants.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+DTYPE_TO_FP8_MAX = {DTypes.FP8E4M3: 448.0, DTypes.FP8E5M2: 57344.0, DTypes.kfloat16: 65504.0}
+
+
+FP8_LINEAR_META_MAPPING = {
+    "input": {"dtype": DTypes.FP8E4M3},
+    "weight": {"dtype": "float16", "fp8_max": 65504.0},
+    "weight_gradient": {"dtype": DTypes.FP8E4M3, "fp8_max": 448.0},
+    "output_gradient": {"dtype": DTypes.FP8E5M2, "fp8_amax": 57344.0},
```
NouamaneTazi, line 33: let's keep hardcoded values in a single place

src/nanotron/fp8/kernel.py
```diff
@@ -0,0 +1,68 @@
+import torch
+import transformer_engine as te  # noqa
```
NouamaneTazi, line 2: Should we add try except? Or add it as a dependency?

src/nanotron/fp8/kernel.py
```diff
@@ -0,0 +1,68 @@
...
+def fp8_matmul_kernel(
+    mat_a: FP8Tensor,
+    transpose_a: bool,
+    mat_b: FP8Tensor,
+    transpose_b: bool,
+    bias: FP8Tensor = None,
+    transpose_bias: bool = False,
+    use_split_accumulator: bool = None,
+) -> torch.Tensor:
+    assert use_split_accumulator is not None
```
NouamaneTazi, line 18: that means we should have `use_split_accumulator` as a required arg?

src/nanotron/fp8/utils.py
```diff
@@ -0,0 +1,13 @@
...
+import transformer_engine as te  # noqa
+
+from nanotron.fp8.constants import FP8_GPU_NAMES
+
+
+def is_fp8_available() -> bool:
+    """Check if FP8 is available on the current device."""
+    if torch.cuda.is_available():
+        device_name = torch.cuda.get_device_name(torch.cuda.current_device()).lower()
+        return any(gpu_name in device_name for gpu_name in FP8_GPU_NAMES)
```
NouamaneTazi, line 11: I'm surprised TE doesn't have an util to check this. Did you search for it?

tests/fp8/test_fp8_parameter.py
```diff
@@ -0,0 +1,25 @@
...
+
+# @pytest.mark.parametrize("quantize_input", [True, False])
+def test_create_fp8_parameter():
+    tensor = torch.randn(16, 16, device="cuda", dtype=torch.float32)
+    # tensor = FP8Tensor(tensor, DTypes.FP8E4M3) if quantize_input else tensor
+
+    fp8_parameter = FP8Parameter(tensor, DTypes.FP8E4M3)
+
+    # TODO(xrsrke): are there better ways for testing
+    # if a tensor is quantized or not?
```
NouamaneTazi, line 16: Depending on how you quantize it, you can test the correctness of the quantized value to be exacly what we want

tests/fp8/test_fp8_parameter.py
```diff
@@ -0,0 +1,25 @@
+import torch
+from nanotron.fp8.constants import DTypes
+from nanotron.fp8.meta import FP8Meta
+from nanotron.fp8.parameter import FP8Parameter
+from nanotron.fp8.tensor import FP8Tensor
+
+
+# @pytest.mark.parametrize("quantize_input", [True, False])
```
NouamaneTazi, line 8: can we keep this?

tests/fp8/test_fp8_parameter.py
```diff
@@ -0,0 +1,25 @@
...
+def test_create_fp8_parameter():
+    tensor = torch.randn(16, 16, device="cuda", dtype=torch.float32)
+    # tensor = FP8Tensor(tensor, DTypes.FP8E4M3) if quantize_input else tensor
```
NouamaneTazi, line 11: can we keep this?

tests/fp8/test_fp8_parameter.py
```diff
@@ -0,0 +1,25 @@
...
+    # TODO(xrsrke): are there better ways for testing
+    # if a tensor is quantized or not?
+    assert isinstance(fp8_parameter.data, FP8Tensor)
+    assert fp8_parameter.requires_grad is True
+    assert fp8_parameter.grad is None
+    assert isinstance(fp8_parameter.fp8_meta, FP8Meta)
+    assert isinstance(fp8_parameter.data.fp8_meta, FP8Meta)
+
+
+# TODO(xrsrke): add test for preventing torch autograd do the backward pass
```
NouamaneTazi, line 24: Yes good test as well, to make sure bwd functions as expected

tests/fp8/test_kernel.py
```diff
@@ -0,0 +1,34 @@
...
+
+    output = _fp8_matmul_kernel(
+        mat_a=fp8_input,
+        transpose_a=transpose_a,
+        mat_b=fp8_weight,
+        transpose_b=transpose_b,
+        use_split_accumulator=use_split_accumulator,
+    )
+
+    assert torch.allclose(output, ref_output, 0.1, 0.1)
```
NouamaneTazi, line 34: 0.1 tolerance is a little too big to test anything? Let's try to calculate ref_output in a way to have 0 tolerance

src/nanotron/fp8/tensor.py
```diff
@@ -0,0 +1,81 @@
...
+class FP8Tensor(torch.Tensor):
+    """FP8 Tensor."""
+
+    # TODO(xrsrke): add type hints for fp8_meta
+    def __new__(cls, tensor: torch.Tensor, dtype: DTypes):
+        # TODO(xrsrke): if the tensor is on cpu, then bypass the quantization
+        assert tensor.device != torch.device("cpu"), "FP8Tensor only supports CUDA device"
+        assert isinstance(dtype, DTypes)
+
+        inverse_scale = torch.tensor(1.0, device=tensor.device, dtype=torch.float32)
```
NouamaneTazi, line 20: does this have to be fp32?

src/nanotron/fp8/meta.py
```diff
@@ -0,0 +1,62 @@
...
+
+    amax: Union[int, float]
+
+    # TODO(xrsrke): change to Literal[torch.int8, torch.uint8]
+    dtype: torch.dtype
+    inverse_scale: float = 1.0
+
+    @property
+    def te_dtype(self) -> tex.DType:
+        from nanotron.fp8.tensor import convert_torch_dtype_to_te_dtype
```
NouamaneTazi, line 23: import to top

src/nanotron/fp8/meta.py
```diff
@@ -0,0 +1,62 @@
...
+    def _compute_scaling_factor(self, margin: float = 0) -> torch.Tensor:
+        # Credits: https://github.com/Azure/MS-AMP/blob/d562f0f0bcfc9b712fa0726b73428753ff1300ab/msamp/common/tensor/meta.py#L39
+        # NOTE: seems like python 3.11.5 don't call __post_init__ when using dataclass
+        # so it don't cast [int, float] to torch.Tensor
+        # so we need to manually cast it
+        amax = torch.tensor(self.amax) if not isinstance(self.amax, torch.Tensor) else self.amax
+        scale = torch.tensor(self.scale) if not isinstance(self.scale, torch.Tensor) else self.scale
+
+        # NOTE: calculate the number of bits to shift the exponent
+        ratio = self.fp8_max / amax
```
NouamaneTazi, line 50: Let's make sure these operations are done in the precision we need and want. For example let's assert the precisions of the used variables

src/nanotron/fp8/tensor.py
```diff
@@ -0,0 +1,81 @@
...
+        fp8_meta = FP8Meta(tensor.abs().max().clone(), dtype, inverse_scale)
+
+        if tensor.dtype not in [torch.int8, torch.uint8]:
```
NouamaneTazi, line 23: We should use our defined constants instead of torch.int8

src/nanotron/fp8/tensor.py
```diff
@@ -0,0 +1,81 @@
...
+    # @property
+    # def fp8_meta(self) -> FP8Meta:
+    #     return self._fp8_meta
+
+    # @fp8_meta.setter
+    # def fp8_meta(self, value: FP8Meta):
+    #     self._fp8_meta = value
+
+    # def __repr__(self) -> str:
+    #     return f"FP8Tensor({self}, fp8_meta={self.fp8_meta})"
```
NouamaneTazi, line 43: why remove these? let's try to keep them

src/nanotron/fp8/tensor.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def convert_torch_dtype_to_te_dtype(dtype: torch.dtype) -> tex.DType:
+    # setattr(torch, "fp8e5m2", torch.int8)
+    # setattr(torch, "fp8e4m3", torch.uint8)
```
NouamaneTazi, line 48: doesnt seem like a good idea

src/nanotron/fp8/tensor.py
```diff
@@ -0,0 +1,81 @@
...
+    from nanotron.fp8.constants import DTypes
```
NouamaneTazi, line 49: imports to top

src/nanotron/fp8/tensor.py
```diff
@@ -0,0 +1,81 @@
...
+
+    # NOTE: transformer engine maintains it own dtype mapping
+    # so we need to manually map torch dtypes to te dtypes
+    TORCH_DTYPE_TE_DTYPE_NAME_MAPPING = {
```
NouamaneTazi, line 53: they don't have a mapping in the library themselves? 🥲

src/nanotron/fp8/meta.py
```diff
@@ -0,0 +1,62 @@
...
+        exp = torch.floor(torch.log2(ratio)) - margin
+        sf = torch.round(torch.pow(2, torch.abs(exp)))
+        sf = torch.where(amax > 0.0, sf, scale)
+        sf = torch.where(torch.isfinite(amax), sf, scale)
+        sf = torch.where(exp < 0, 1 / sf, sf)
+        return sf
+
+    def _update_inverse_scale(self):
+        self.inverse_scale = 1 / self.scale
```
NouamaneTazi, line 59: Do we really need to keep track of both variables scale and inverse_scale if we can't get one by inversing the other?
I'd say if it doesn't give us huge optims, then let's get rid of one of them

src/nanotron/fp8/meta.py
```diff
@@ -0,0 +1,62 @@
...
+
+        return convert_torch_dtype_to_te_dtype(self.dtype)
+
+    def __post_init__(self):
+        # NOTE: transformer engine only accepts torch tensors
+        # NOTE: remove fill scale factor when initializing
+        # TODO(xrsrke): check if we haven't computed the scaling factor, then do an initial scaling factor
+        # don't set the initial to 1
+        self.scale = 1  # initial scaling factor
+        self.scale = self._compute_scaling_factor()
```
NouamaneTazi, line 33: Why don't we have `scale` as an attr of `FP8Meta` ?

src/nanotron/fp8/meta.py
```diff
@@ -0,0 +1,62 @@
...
+        """Return the maximum normal value for the current dtype."""
+        return DTYPE_TO_FP8_MAX[self.dtype]
+
+    def _compute_scaling_factor(self, margin: float = 0) -> torch.Tensor:
+        # Credits: https://github.com/Azure/MS-AMP/blob/d562f0f0bcfc9b712fa0726b73428753ff1300ab/msamp/common/tensor/meta.py#L39
+        # NOTE: seems like python 3.11.5 don't call __post_init__ when using dataclass
+        # so it don't cast [int, float] to torch.Tensor
+        # so we need to manually cast it
+        amax = torch.tensor(self.amax) if not isinstance(self.amax, torch.Tensor) else self.amax
+        scale = torch.tensor(self.scale) if not isinstance(self.scale, torch.Tensor) else self.scale
```
NouamaneTazi, line 47: not a big fan of doing under the hood conversions. Let's make sure `amax` and `scale` are converted correctly to tensors when declaring `FP8Meta` and that they stay like that

src/nanotron/fp8/meta.py
```diff
@@ -0,0 +1,62 @@
...
+
+        # NOTE: calculate the number of bits to shift the exponent
+        ratio = self.fp8_max / amax
+        exp = torch.floor(torch.log2(ratio)) - margin
+        sf = torch.round(torch.pow(2, torch.abs(exp)))
+        sf = torch.where(amax > 0.0, sf, scale)
+        sf = torch.where(torch.isfinite(amax), sf, scale)
+        sf = torch.where(exp < 0, 1 / sf, sf)
```
NouamaneTazi, line 55: hoping we don't call this function often bcz these are a lot of kernels 🥲

src/nanotron/fp8/meta.py
```diff
@@ -0,0 +1,62 @@
...
+        self.scale = 1  # initial scaling factor
+        self.scale = self._compute_scaling_factor()
+        self.amax = torch.tensor(self.amax, device="cuda") if not isinstance(self.amax, torch.Tensor) else self.amax
+
+    @property
+    def fp8_max(self) -> float:
+        """Return the maximum normal value for the current dtype."""
+        return DTYPE_TO_FP8_MAX[self.dtype]
+
+    def _compute_scaling_factor(self, margin: float = 0) -> torch.Tensor:
```
NouamaneTazi, line 41: can we torch.jit.script this func?

src/nanotron/fp8/meta.py
NouamaneTazi, line 47: Not sure I understand why we need previous `scale` to compute the scaling factor?

NouamaneTazi: maybe better to rename func to `update` instead of `compute`?

src/nanotron/fp8/kernel.py
```diff
@@ -0,0 +1,68 @@
...
+from nanotron.fp8.tensor import FP8Tensor
+
+
+@torch.no_grad()
+def fp8_matmul_kernel(
+    mat_a: FP8Tensor,
+    transpose_a: bool,
+    mat_b: FP8Tensor,
+    transpose_b: bool,
+    bias: FP8Tensor = None,
```
NouamaneTazi, line 14: add warning that bias is not supported yet

src/nanotron/fp8/parameter.py
```diff
@@ -0,0 +1,37 @@
...
+    @data.setter
+    def data(self, data):
+        self._data = data
+
+    @property
+    def fp8_meta(self):
+        return self.data.fp8_meta
+
+    # def __repr__(self) -> str:
+    #     return f"FP8Parameter({self.data}, fp8_meta={self.fp8_meta}, requires_grad={self.requires_grad}"
```
NouamaneTazi, line 37: always nice to keep parent's class's repr

src/nanotron/fp8/linear.py
```diff
@@ -0,0 +1,107 @@
...
+        # because weight and bias's requires grad are set to False
+        # so that we can compute the gradients using fp8 kernels by ourselves
+        phony = torch.empty(0, device=input.device, requires_grad=True)
+        output, _ = _FP8MatMul.apply(input, self.weight, self.fp8_meta, phony)
+
+        output = output if self.bias is None else output + self.bias
+        return output
+
+
+class _FP8MatMul(torch.autograd.Function):
```
NouamaneTazi, line 55: nit: rename `MatMul` to `Matmul`

src/nanotron/fp8/linear.py
```diff
@@ -0,0 +1,107 @@
...
+    @staticmethod
+    def forward(
+        ctx, input: FP8Tensor, weight: FP8Tensor, fp8_meta: FP8LinearMeta, phony: torch.Tensor
+    ) -> torch.Tensor:
+        with torch.no_grad():
```
NouamaneTazi, line 60: decorate all of fwd with no_grad then?

NouamaneTazi (changes requested):
Amazing job so far!! This is gonna be lit 🔥🔥🔥🔥

src/nanotron/fp8/constants.py
xrsrke: yes!!

tests/fp8/test_fp8_parameter.py
xrsrke: will do it in another PR 🥲

src/nanotron/fp8/tensor.py
xrsrke: will do it in another PR 🥲

tests/fp8/test_fp8_parameter.py
xrsrke: there are a few issues. will fix it in another PR!! 

https://github.com/huggingface/nanotron/blob/715bf38dd2c82f136e94e25f7e32e440bee7967c/src/nanotron/fp8/parameter.py#L22

src/nanotron/fp8/kernel.py
xrsrke: Hmm, maybe we should use try-except. That way, we don't complicate things for people who just want to try out Nanotron (without having to install TE in order to try non-FP8 features...

tests/fp8/test_kernel.py
xrsrke: In the FP8-LM paper, they also use 0.1 as a tolerance, and they show that the convergence is acceptable for 125M, 7B, 13B, and 175B models!

xrsrke: Oops. this is duplication test. We already have it here: https://github.com/huggingface/nanotron/blob/1702d21554d1c9fd2046e057d55356bac788e165/tests/fp8/test_linear.py#L36

src/nanotron/fp8/tensor.py
xrsrke: They have this https://github.com/NVIDIA/TransformerEngine/blob/d68028c872186cd3df604e2ee2f60e09784d955c/transformer_engine/pytorch/cpp_extensions/fused_attn.py#L26. But it looks like a low-level detail. I'm afraid that if we depend on this and they refactor it out (since it's a low-level implementation detail), our code could break!!

src/nanotron/fp8/utils.py
xrsrke: sadly nope :) https://github.com/search?q=repo%3ANVIDIA%2FTransformerEngine%20torch.cuda.get_device_name&type=code

src/nanotron/fp8/meta.py
xrsrke: added!

thomwolf (commented):
Looks good to me. Small nit that I would just move the `warnings.warn("Please install Transformer engine for FP8 training!")` in either a selective import in the `__init__.py` or a decorator testing if `te is None` before the various FP8 functions to avoid having the warnings all  the time

src/nanotron/fp8/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+from nanotron.fp8.dtypes import DTypes  # noqa
+from nanotron.fp8.linear import FP8Linear  # noqa
+from nanotron.fp8.parameter import FP8Parameter  # noqa
+from nanotron.fp8.tensor import FP8Tensor  # noqa
+
+try:
+    import transformer_engine as te  # noqa
+    import transformer_engine_extensions as tex  # noqa
+except ImportError:
+    warnings.warn("Please install Transformer engine for FP8 training!")
```
NouamaneTazi, line 12: add pip install instruction

src/nanotron/fp8/kernel.py
NouamaneTazi: try except still missing no?

src/nanotron/fp8/kernel.py
```diff
@@ -0,0 +1,69 @@
...
+    mat_a_fp8_meta = mat_a.fp8_meta
+    mat_b_fp8_meta = mat_b.fp8_meta
+
+    # NOTE: these are the fixed configs that TE only takes
+    # so we have to TE the A and B matrix to match these configs
+    TE_CONFIG_TRANSPOSE_A = True
+    TE_CONFIG_TRANSPOSE_B = False
+    SCALE = AMAX = _empty_tensor
+
+    mat_a = tex.fp8_transpose(mat_a, mat_a_fp8_meta.te_dtype) if transpose_a is False else mat_a
```
NouamaneTazi, line 42: do we have to do these transposes? can't we just use the args `TE_CONFIG_TRANSPOSE_A` ?

src/nanotron/fp8/linear.py
```diff
@@ -0,0 +1,112 @@
...
+class FP8Linear(nn.Linear):
+    def __init__(self, in_features: int, out_features: int, bias: bool = True, device: Optional[torch.device] = None):
+        super().__init__(in_features, out_features, bias, device)
+        # TODO(xrsrke): add device, and 2 fp8 dtypes
+        if self.weight.device != torch.device("cpu"):
+            self.weight = FP8Parameter(self.weight, dtype=DTypes.FP8E4M3)
+
+            # NOTE: quantization metadata for input gradients, weight gradients, and output gradients
+            # TODO(xrsrke): don't fixed this
+            FP8E4M3_SCALE = update_scaling_factor(
```
NouamaneTazi, line 33: this variable can be lowercase

src/nanotron/fp8/linear.py
```diff
@@ -0,0 +1,112 @@
...
+                amax=torch.tensor(INITIAL_AMAX, dtype=torch.float32),
+                scaling_factor=torch.tensor(INITIAL_SCALING_FACTOR),
+                dtype=DTypes.FP8E4M3,
+            )
+            FP8E5M2_SCALE = update_scaling_factor(
```
NouamaneTazi, line 38: same

src/nanotron/fp8/linear.py
```diff
@@ -0,0 +1,112 @@
...
+    def __init__(self, in_features: int, out_features: int, bias: bool = True, device: Optional[torch.device] = None):
+        super().__init__(in_features, out_features, bias, device)
+        # TODO(xrsrke): add device, and 2 fp8 dtypes
+        if self.weight.device != torch.device("cpu"):
+            self.weight = FP8Parameter(self.weight, dtype=DTypes.FP8E4M3)
+
+            # NOTE: quantization metadata for input gradients, weight gradients, and output gradients
+            # TODO(xrsrke): don't fixed this
+            FP8E4M3_SCALE = update_scaling_factor(
+                amax=torch.tensor(INITIAL_AMAX, dtype=torch.float32),
```
NouamaneTazi, line 34: can you use DTypeInvariantTensor to make sure this tensor stays in float32? See this context manager that we always use when building models: https://github.com/huggingface/nanotron/blob/b1c26750b2b541efa668fb9f9d04abe24a4939ae/src/nanotron/models/base.py#L206-L209

src/nanotron/fp8/kernel.py
```diff
@@ -0,0 +1,69 @@
...
+    accumulate = False
+
+    out_dtype = getattr(tex.DType, "kFloat32")
+    # NOTE: currently TE don't support adding bias in FP8
+    # along with matmul, it only takes an empty bias
+    bias = torch.tensor([], dtype=torch.float32)
+    TE_CONFIG_TRANSPOSE_BIAS = False
+
+    mat_a_fp8_meta = mat_a.fp8_meta
+    mat_b_fp8_meta = mat_b.fp8_meta
```
NouamaneTazi, line 34: add typing to these variables

src/nanotron/fp8/linear.py
```diff
@@ -0,0 +1,112 @@
...
+    @staticmethod
+    @torch.no_grad()
+    def forward(
+        ctx, input: FP8Tensor, weight: FP8Tensor, fp8_meta: FP8LinearMeta, phony: torch.Tensor
+    ) -> torch.Tensor:
+        if type(input) == torch.Tensor:
+            input = FP8Tensor(input, dtype=DTypes.FP8E4M3)
+
+        ctx.save_for_backward(input, weight)
+        ctx.fp8_meta = fp8_meta
```
NouamaneTazi, line 79: I'm not sure what's the point of having fp8_meta here? if we also have fp8_meta in `weight`?

src/nanotron/fp8/linear.py
```diff
@@ -0,0 +1,112 @@
...
+        ∂L/∂W = Xᵀ @ ∂L/∂Y
+        Source: https://web.eecs.umich.edu/~justincj/teaching/eecs442/notes/linear-backprop.html
+        """
+        # TODO(xrsrke): investigate how does grad_output.contiguous() affect the outputs
+        input, weight = ctx.saved_tensors
+
+        if type(grad_output) == torch.Tensor:
+            grad_output = torch.ones_like(grad_output)
+            grad_output = grad_output.contiguous()
+            grad_output = FP8Tensor(grad_output, dtype=DTypes.FP8E5M2)
```
NouamaneTazi, line 102: shouldn't this be using fp8_meta for grad? can we/should we assume that fp8_meta for grad is same as for weights?

src/nanotron/fp8/linear.py
```diff
@@ -0,0 +1,112 @@
...
+                scaling_factor=torch.tensor(INITIAL_SCALING_FACTOR, dtype=torch.float32),
+                dtype=DTypes.FP8E5M2,
+            )
+            self.fp8_meta: FP8LinearMeta = {
+                # kfloat8_e4m3
+                "input_grad": FP8Meta(amax=1, dtype=DTypes.FP8E4M3, scale=FP8E4M3_SCALE),
+                "weight_grad": FP8Meta(amax=1, dtype=DTypes.FP8E4M3, scale=FP8E4M3_SCALE),
+                # kfloat8_e5m2
+                "output_grad": FP8Meta(amax=1, dtype=DTypes.FP8E5M2, scale=FP8E5M2_SCALE),
+            }
```
NouamaneTazi, line 49: this block seems to be constant. Why are we creating this dict at init time of the Linear?

src/nanotron/fp8/parameter.py
```diff
@@ -0,0 +1,42 @@
...
+        assert data.dtype not in FP8_DTYPES, "Currently only support turn a non-fp8 tensor to an fp8 parameter"
+        assert data.device != torch.device("cpu"), "FP8Parameter only supports CUDA tensors"
+        # TODO(xrsrke): if the tensor is on cpu, then bypass quantization
+
+        with torch.no_grad():
+            # TODO(xrsrke): support take an FP8 Tensor as data
+            # currently we can't only quantize a tensor to FP8 after the parameter is created
+            # because it raise "Only Tensors of floating point and complex dtype can require gradients"
+            self = torch.Tensor._make_subclass(cls, data, requires_grad)
+            self._data = FP8Tensor(data, dtype=dtype) if isinstance(data, torch.Tensor) else data
```
NouamaneTazi, line 26: I think we'll be less error prone if we assert that data is of instance `torch.Tensor` just like the typing of this function says

src/nanotron/fp8/linear.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+class _FP8Matmul(torch.autograd.Function):
+    @staticmethod
+    @torch.no_grad()
+    def forward(
+        ctx, input: FP8Tensor, weight: FP8Tensor, fp8_meta: FP8LinearMeta, phony: torch.Tensor
+    ) -> torch.Tensor:
+        if type(input) == torch.Tensor:
+            input = FP8Tensor(input, dtype=DTypes.FP8E4M3)
```
NouamaneTazi, line 76: I think it's less error prone if we assert input's type to be `FP8Tensor`
Or if we want to keep this liberty update the typing in the function

src/nanotron/fp8/tensor.py
```diff
@@ -0,0 +1,106 @@
...
+from nanotron.fp8.dtypes import DTypes
+
+
+class FP8Tensor(torch.Tensor):
+    """FP8 Tensor."""
+
+    def __new__(cls, tensor: torch.Tensor, dtype: DTypes) -> torch.Tensor:
+        # TODO(xrsrke): there is a circular import issue
+        # between tensor.py and meta.py fix this
+        from nanotron.fp8.meta import FP8Meta
```
NouamaneTazi, line 15: try to make one file import from the other? can we move convert_torch_dtype_to_te_dtype to meta.py?

src/nanotron/fp8/tensor.py
```diff
@@ -0,0 +1,106 @@
...
+        assert isinstance(dtype, DTypes)
+
+        amax = tensor.abs().max().clone()
+        scale = update_scaling_factor(amax, torch.tensor(INITIAL_SCALING_FACTOR, dtype=torch.float32), dtype)
+        fp8_meta = FP8Meta(amax, scale, dtype)
+
+        if tensor.dtype not in FP8_DTYPES:
+            fp8_tensor = convert_tensor_to_fp8(tensor, fp8_meta)
+        else:
+            fp8_tensor = tensor
```
NouamaneTazi, line 29: what does it mean to have `tensor` alrdy in fp8. I think that may create issues because we're overriding its fp8_meta. Let's assert `tensor` is not fp8

src/nanotron/fp8/tensor.py
```diff
@@ -0,0 +1,106 @@
...
+
+        # TODO(xrsrke): move update inverse scaling to FP8Meta's initialization
+        obj = torch.Tensor._make_subclass(cls, fp8_tensor)
+        obj.fp8_meta = fp8_meta
+        return obj
+
+    def __repr__(self) -> str:
+        return f"FP8Tensor({self}, fp8_meta={self.fp8_meta})"
```
NouamaneTazi, line 37: I like these!

src/nanotron/fp8/tensor.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+def update_scaling_factor(
+    amax: torch.Tensor, scaling_factor: torch.Tensor, dtype: DTypes, margin: float = 0
+) -> torch.Tensor:
+    """
+    Update the scaling factor to quantize a tensor to FP8.
+    Credits: https://github.com/Azure/MS-AMP/blob/d562f0f0bcfc9b712fa0726b73428753ff1300ab/msamp/common/tensor/meta.py#L39
+    """
+    assert amax.dtype == torch.float32
```
NouamaneTazi, line 85: nice!

src/nanotron/fp8/tensor.py
```diff
@@ -0,0 +1,106 @@
...
+    assert scaling_factor.dtype == torch.float32
+
+    # NOTE: Since fp8_max is a fixed number based on two FP8 data types,
+    # we prefer not to take fp8_max in the input arguments.
+    fp8_max = torch.tensor(DTYPE_TO_FP8_MAX[dtype], dtype=torch.float32)
+
+    # NOTE: torch.jit only take a concrete value rather than a DTYPE_TO_FP8_MAX[dtype],
+    # so we create an inner function to bypass that
+    @torch.jit.script
+    def _inner(amax: torch.Tensor, fp8_max: torch.Tensor, scaling_factor: torch.Tensor, margin: float):
```
NouamaneTazi, line 96: very cool!

tests/fp8/test_linear.py
```diff
@@ -0,0 +1,83 @@
...
+    if is_bias:
+        fp8_linear.bias.data = ref_linear.bias.detach().clone()
+
+    ref_output = ref_linear(ref_input)
+    output = fp8_linear(input)
+
+    assert isinstance(output, torch.Tensor)
+    assert output.dtype == torch.float32
+    assert torch.allclose(output, ref_output, rtol=0, atol=0.1)
+
```
NouamaneTazi, line 26: can we have stricter comparisons?
also u can use `torch.testing.assert_allclose`

tests/fp8/test_linear.py
```diff
@@ -0,0 +1,83 @@
...
+        fp8_linear.weight.data = FP8Tensor(ref_linear.weight.detach().clone(), dtype=DTypes.FP8E4M3)
+    fp8_linear.bias.data = ref_linear.bias.detach().clone()
+
+    ref_linear(ref_input).sum().backward()
+    fp8_linear(input).sum().backward()
+
+    # TODO(xrsrke): investigate why input.grad is so high tolerance
+    # assert torch.allclose(input.grad, ref_input.grad, 0.2, 0.2) if input_requires_grad else True
+    assert torch.allclose(fp8_linear.weight.grad, ref_linear.weight.grad, 0.1, 0.1)
+    assert torch.allclose(fp8_linear.bias.grad, ref_linear.bias.grad, 0, 0.1)
```
NouamaneTazi, line 49: need stricter comparisons

tests/fp8/test_linear.py
```diff
@@ -0,0 +1,83 @@
...
+        *[nn.Sequential(FP8Linear(HIDEEN_SIZE, HIDEEN_SIZE, device="cuda"), nn.ReLU()) for _ in range(N_LAYERS)]
+    )
+    optim = Adam(model.parameters(), lr=1e-3)
+
+    for _ in range(N_EPOCHS):
+        optim.zero_grad()
+        model(input).sum().backward()
+        optim.step()
+
+    assert all(p.grad is not None for p in model.parameters())
```
NouamaneTazi, line 83: what are we testing here?

NouamaneTazi (changes requested):

tests/fp8/test_linear.py
xrsrke: I use the tolerance that FP8-LM uses, and with this tolerance, the paper show that there is a minimal degradation in convergence => it's fine

image

tests/fp8/test_linear.py
xrsrke: Ah, this PR is only for a single iteration of forward and backward; will update this test in the next PR, which will support multiple iterations

tests/fp8/test_linear.py
xrsrke: > can we have stricter comparisons?
i use the tolerance from msamp [[link]](https://github.com/Azure/MS-AMP/blob/ac666ace7dec72b8b6b6e8ad47484869a09ca18a/tests/nn/test_linear.py#L39), the paper shows that this is good enough

src/nanotron/fp8/kernel.py
xrsrke: Yes, I tried it, but TE raises errors (something like it's not supported); it doesn't support transposing tensor A. We have to transpose the inputs in a way that we will obtain an equivalent output before feeding it to the kernel

src/nanotron/fp8/linear.py
xrsrke: > can we/should we assume that fp8_meta for grad is same as for weights?
Yes, i'll refactor out the weight gradient's metadata to the gradient tensor itself

src/nanotron/fp8/linear.py
xrsrke: because the metadata will changing during training => it bounds to instance of a fp8linear

src/nanotron/fp8/linear.py
xrsrke: I want to make it allow taking an input in FP8 format. Should we remove that?

src/nanotron/fp8/tensor.py
xrsrke: will do in another PR 🫠

src/nanotron/fp8/tensor.py
xrsrke: good catch. thanks

src/nanotron/fp8/linear.py
xrsrke: A weight ideally should only have the metadata of its data and its gradients, but in a linear layer, we also need metadata for output gradients. The reason you see it's not being used yet is because this PR only supports FP8 for a single iteration (no optim.step()). Will update it in the next PR, which allows multiple .steps => gradients => updating gradient metadata

src/nanotron/fp8/kernel.py
```diff
@@ -0,0 +1,75 @@
...
+
+import torch
+
+from nanotron.fp8.tensor import FP8Tensor
+
+try:
+    import transformer_engine as te  # noqa
+    import transformer_engine_extensions as tex
+except ImportError:
+    warnings.warn("Please install Transformer engine for FP8 training!")
```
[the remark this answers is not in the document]
xrsrke: Maybe we should expect users to check how to install it themselves because how to install TE could change over time?

src/nanotron/fp8/__init__.py
xrsrke: same as above

src/nanotron/fp8/linear.py
xrsrke: will do it in the next PR 🫠

src/nanotron/fp8/kernel.py
xrsrke: these are the low-level files, we expect users to call FP8Tensor, FP8Linear, or Dtypes only (from nanotron.fp8) => we could skip it?

NouamaneTazi (approved):
Noiceee! I've noticed some comments are still not addressed but looking forward to seeing them in the next PR!!

PR #58, opened 2024-02-06
Adding eval config

thomwolf:
support for lightevals

### DIFF
--- a/src/nanotron/config/config.py
+++ b/src/nanotron/config/config.py
@@ -337,21 +337,23 @@ class Config:
 
 
 def get_config_from_dict(
-    config_dict: dict, config_class: Type = Config, skip_unused_config_keys: bool = True, skip_null_keys: bool = True
+    config_dict: dict, config_class: Type = Config, skip_unused_config_keys: bool = False, skip_null_keys: bool = False
 ):
     """Get a config object from a dictionary
 
     Args:
         args: dictionary of arguments
         config_class: type of the config object to get as a ConfigTypes (Config, LightevalConfig, LightevalSlurm) or str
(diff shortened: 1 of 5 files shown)


src/nanotron/config/config.py
```diff
@@ -375,10 +336,30 @@ def as_dict(self) -> dict:
         return serialize(self)
 
 
-def get_config_from_dict(args: dict, config_class: Type[Config] = Config):
+def get_config_from_dict(
+    config_dict: dict, config_class: Type = Config, skip_unused_config_keys: bool = True, skip_null_keys: bool = True
```
NouamaneTazi, line 340: maybe we want to default the skips to False?

src/nanotron/config/config.py
```diff
@@ -375,10 +336,30 @@ def as_dict(self) -> dict:
...
+        args: dictionary of arguments
+        config_class: type of the config object to get as a ConfigTypes (Config, LightevalConfig, LightevalSlurm) or str
+        skip_unused_config_keys: whether to skip unused first-level keys in the config file (for config with additional sections)
+        skip_null_keys: whether to skip keys with value None at first and second level
+    """
+    if skip_unused_config_keys:
+        config_dict = {
+            field.name: config_dict[field.name] for field in fields(config_class) if field.name in config_dict
+        }
+    if skip_null_keys:
```
NouamaneTazi, line 354: should we add warnings when we skip keys or no need to?

NouamaneTazi (approved):
LGTM modulo minor comments

src/nanotron/config/config.py
thomwolf: yes sounds good

PR #68, opened 2024-02-13
Fix topology agnostic loading

nopperl:
When loading a checkpoint with a different tp degree from the configured tp degree, the following error is raised:

```
Traceback (most recent call last):
  File "/home/nanotron/run_train.py", line 132, in <module>
    trainer = DistributedTrainer(config_file)
  File "/home/nanotron/src/nanotron/trainer.py", line 162, in __init__
    load_optimizer(
  File "/home/conda/envs/linux/lib/python3.10/site-packages/torch/utils/_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "/home/nanotron/src/nanotron/serialize/optimizer.py", line 222, in load_optimizer
    ckp_shard_data = ckp_optim_state["state"][optim_state_index][state_key]
KeyError: None
```

This happens only for the `model.lm_head.pp_block.weight` parameter. I assume this is because the optimizer states for this parameter are stored under the tied `model.token_position_embeddings.pp_block.token_embedding.weight` parameter. This PR fixes this by skipping trying to load the lm_head optimizer states. This is similar to weight loading, where the `model.token_position_embeddings.pp_block.token_embedding.weight` weights are loaded for `model.lm_head.pp_block.weight` (see https://github.com/huggingface/nanotron/blob/main/src/nanotron/serialize/weights.py#L347), but I think the optimizer states can be skipped.

To reproduce:

Setup config files:

```
cat > examples/debug_topology_agnostic.yaml << EOL
# CUDA_DEVICE_MAX_CONNECTIONS=1 torchrun --nproc_per_node=4 run_train.py --config-file examples/debug_topology_agnostic.yaml

checkpoints:
  checkpoint_interval: 10
  checkpoints_path: checkpoints/debug_topology_agnostic
  checkpoints_path_is_shared_file_system: true
  save_initial_state: false
data:
  dataset:
  num_loading_workers: 1
  seed: 42
general:
  benchmark_csv_path: null
  consumed_train_samples: null
  ignore_sanity_checks: false
  project: debug
  run: tiny_llama
  seed: 42
  step: null
logging:
  iteration_step_info_interval: 1
  log_level: info
  log_level_replica: info
model:
  ddp_bucket_cap_mb: 25
  dtype: float16
  init_method:
    std: 0.025
  make_vocab_size_divisible_by: 1
  model_config:
    bos_token_id: 1
    eos_token_id: 2
    hidden_act: silu
    hidden_size: 32
    initializer_range: 0.02
[...]

Issue #38, opened 2024-01-23
Merging optimizer states from different pipeline parallel size to resume training


xrsrke:
Suppose you start training with a pipeline parallel size of 4. We need to make it supports resuming training with a different pipeline parallel size, like 2, by merging optimizer states.

xrsrke:
Suppose you start training with a pipeline parallel size of 4. We need to make it supports resuming training with a different pipeline parallel size, like 2, by merging optimizer states.

PR #71, opened 2024-02-15
Implement pipeline parallel size-agnostic optimizer state loading

nopperl:
I tried implementing the topology-agnostic optimizer state loading for the pipeline parallel dimension ( #38 ). This also fixes the issue I had in #68 .


NouamaneTazi (approved):
LGTM!! Very clean PR. Thanks a lot

PR #72, opened 2024-02-16
Support Expert Parallelism

NouamaneTazi:


src/nanotron/config/parallelism_config.py
```diff
@@ -20,6 +20,7 @@ class ParallelismArgs:
         dp: Number of DP replicas
         pp: Number of PP stages
         tp: Number of TP replicas
+        expert_parallel_size: Number of expert parallel replicas (used only for MoEs)
```
xrsrke, line 23: Isn't expert_parallel_size should be the number of experts per tp rank?

src/nanotron/config/parallelism_config.py
```diff
@@ -34,6 +35,8 @@ class ParallelismArgs:
     recompute_granularity: Optional[RecomputeGranularity] = None
     tp_linear_async_communication: Optional[bool] = None
 
+    expert_parallel_size: Optional[int] = None
```
xrsrke, line 38: 1 by default?

src/nanotron/parallel/context.py
```diff
@@ -16,9 +16,10 @@ def __init__(
         pipeline_parallel_size: int,
         data_parallel_size: int,
         backend: DistributedBackend = "nccl",
+        expert_parallel_size: int = 1,
```
xrsrke, line 19: Move it after `data_parallel_size` (parallel_size..., backend)

src/nanotron/trainer.py
```diff
@@ -198,7 +199,7 @@ def __init__(
 
         # Setup tensorboard write and log writers on output rank
         self.logger_ranks = self.parallel_context.world_rank_matrix[
-            self.unwrapped_model.output_pp_rank, 0, 0
+            0, self.unwrapped_model.output_pp_rank, 0, 0
```
xrsrke, line 202: would be nice if we do self.parallel_context.world_rank_matrix(expert_rank=0, pp_rank=0, tp_rank=0, dp_rank=0)

src/nanotron/trainer.py
```diff
@@ -350,17 +351,8 @@ def training_step(
...
-                mp_pg=self.parallel_context.world_ranks_to_pg[
-                    tuple(
-                        sorted(
-                            self.parallel_context.world_rank_matrix[
-                                :, dist.get_rank(self.parallel_context.dp_pg), :
-                            ].reshape(-1)
-                        )
-                    )
-                ],
+                mp_pg=self.parallel_context.mp_pg,
```
xrsrke, line 355: nice!

xrsrke (changes requested):

src/nanotron/config/parallelism_config.py
NouamaneTazi: Not quite, expert parallelism is orthogonal to TP. for example you can have 1 expert sharded along 2 tp ranks

src/nanotron/config/parallelism_config.py
NouamaneTazi: true

src/nanotron/trainer.py
NouamaneTazi: i very much like this! but that could be in an othogonal PR

NouamaneTazi: added a TODO

PR #80, opened 2024-02-25
[Feature] Add loading different datasets based on training stages

xrsrke:
Reproduce
- Step 1: Modify your config:

**Use a single dataset for the entire training**
```
data:
  dataset:
      dataset_overwrite_cache: false
      dataset_processing_num_proc_per_process: 1
      hf_dataset_config_name: null
      hf_dataset_or_datasets: HuggingFaceH4/testing_alpaca_small
      hf_dataset_splits: train
      text_column_name: completion
```

**Use different datasets based on training stages**

```
  # NOTE: if you wanna use different datasets for different stages of the training
data:
  dataset_stages:
    - name: Stable Training Stage
      training_steps: 1
      dataset:
        dataset_overwrite_cache: false
        dataset_processing_num_proc_per_process: 1
        hf_dataset_config_name: null
        hf_dataset_or_datasets: HuggingFaceH4/testing_alpaca_small
        hf_dataset_splits: train
        text_column_name: completion
    - name: Annealing Phase
      training_steps: 10
      dataset:
        dataset_overwrite_cache: false
        dataset_processing_num_proc_per_process: 1
        hf_dataset_config_name: null
        hf_dataset_or_datasets: stas/c4-en-10k
        hf_dataset_splits: train
        text_column_name: text
```

- Step 2: `CUDA_DEVICE_MAX_CONNECTIONS=1 torchrun --nproc_per_node=8 run_train.py --config-file examples/config_tiny_llama.yaml`


run_train.py
```diff
@@ -117,11 +201,28 @@ def get_dataloader(trainer: DistributedTrainer):
...
+        sorted_stages = sorted(trainer.config.data.dataset_stages, key=lambda stage: stage.training_steps)
+        dataloaders = {}
+        for idx, stage in enumerate(sorted_stages):
+            dataloader = (
+                get_dataloader(trainer, stage.dataset)
+                if idx == 0
+                else lambda stage=stage: get_dataloader(trainer, stage.dataset)
+            )
+            dataloaders[stage.name] = dataloader
+        return dataloaders
```
NouamaneTazi, line 223: How does this compare with having a single dataloader (so no need to change our trainer dataloader logic)?

NouamaneTazi: - CPU/Memory: does allocating n dataloaders mean we allocate n times the number of workers?
- Determinism: Are we sure we keep determinism when reloading? Let’s say we remove the 1st training stage, are we sure the dataloader will still feed the same data?

NouamaneTazi (commented):
Great job! Left some questiosn

src/nanotron/trainer.py
```diff
@@ -299,12 +312,20 @@ def train(
         torch.cuda.empty_cache()
         with prof:
             for self.iteration_step in range(self.start_iteration_step + 1, self.config.tokens.train_steps + 1):
+                # NOTE: the second condition is for if we have a single training stages
+                # and this is the first iteration step
+                if self.config.data.dataset_stages is not None or (
+                    self.config.data.dataset_stages is None and self.iteration_step == 1
+                ):
+                    self._update_dataloader_based_on_training_stages(dataloader_or_dls)
```
NouamaneTazi, line 320: maybe to keep the `train` code simpler, we can move the if logic inside `self._update_dataloader_based_on_training_stages`

run_train.py
xrsrke: > Are we sure we keep determinism when reloading? Let’s say we remove the 1st training stage, are we sure the dataloader will still feed the same data?
the second dataloader use a different dataset, it use the same seed in the data section, so as long as the another run use the same seed, it should the same (same for the first dataloader)

xrsrke: > does allocating n dataloaders mean we allocate n times the number of workers?
will add remove the earlier dataloaders... and destroy its workers...

Issue #77, opened 2024-02-19
[Feature] Refactor `ParallelContext.world_rank_matrix`


NouamaneTazi:
For now we're storing the global ranks inside the `world_rank_matrix` attribute which is a numpy array of shape `(expert_parallel_size, pipeline_parallel_size, data_parallel_size, tensor_parallel_size)`

So in order to access a process' global rank using the world_rank_matrix right now we're using:
https://github.com/huggingface/nanotron/blob/7c01d0f03dff537bbec79a380d861cb9934ba583/tests/test_parameters_accumulate_gradient_in_fp32.py#L346-L351

It would be cool to make it a functional call instead such as:
```python
parallel_context.get_global_rank(expert_parallel_rank=0, pipeline_parallel_rank=0, data_parallel_rank=0, tensor_parallel_rank=0)
```

info-gallary:
Hi! I’d like to work on this issue if it’s still available. I can implement the "get_global_rank()" helper and add/update the relevant tests. Please let me know if I should proceed.

PR #94, opened 2024-03-03
[Feature] Refactor ParallelContext.world_rank_matrix

0xkerem:
Hi,

I've addressed Issue #77 by introducing the `get_global_rank` method instead of direct access to the `world_rank_matrix` attribute, as suggested. This change enhances code modularity and adheres to best practices.

I've thoroughly tested the modifications to ensure stability. Your feedback is appreciated.

```diff
- parallel_context.world_rank_matrix[
-     dist.get_rank(parallel_context.expert_pg),
-     get_pp_rank_of(target, module=mdl),
-     dist.get_rank(parallel_context.dp_pg),
-     dist.get_rank(parallel_context.tp_pg),
- ],
+ parallel_context.get_global_rank(
+     expert_parallel_rank=dist.get_rank(parallel_context.expert_pg),
+     pipeline_parallel_rank=get_pp_rank_of(target, module=mdl),
+     data_parallel_rank=dist.get_rank(parallel_context.dp_pg),
+     tensor_parallel_rank=dist.get_rank(parallel_context.tp_pg),
+ ),
```


xrsrke:
@NouamaneTazi any thoughts?

xrsrke:
@0xkerem very nice. Thanks for the PR. could you write a unit test for this please?

tests/test_parameters_accumulate_gradient_in_fp32.py
```diff
@@ -343,12 +343,12 @@ def _test_tied_weights_sync_with_grad_accum_in_fp32(
                 (
                     target,
                     (
-                        parallel_context.world_rank_matrix[
-                            dist.get_rank(parallel_context.expert_pg),
-                            get_pp_rank_of(target, module=mdl),
-                            dist.get_rank(parallel_context.dp_pg),
-                            dist.get_rank(parallel_context.tp_pg),
-                        ],
+                        parallel_context.get_global_rank(
```
NouamaneTazi, line 346: There are two more occurences of `world_rank_matrix` in `src/nanotron/trainer.py` can you modify those as well?

NouamaneTazi (changes requested):
looks much cleaner thanks!

tests/test_parameters_accumulate_gradient_in_fp32.py
0xkerem: Hello, thank you very much for your response! I have committed these changes, now I will see if I can create a unit test for it.

xrsrke: @0xkerem you could use this as a reference: https://github.com/huggingface/nanotron/blob/d059fcae1d58da30bb2d6672fb9d8982dc0368be/tests/test_distributed.py#L14-L42

0xkerem: @xrsrke I've been trying to solve an error since yesterday but I couldn't succeed. This error is not caused by nanotron, I don't want to be misunderstood. It's something from CUDA. It started after an update but I haven't figured it out yet. I removed all the Python environment and libraries and reinstalled them to solve the errors caused by the versions. At first, `torch.cuda.is_available()` was returning false. I fixed this but for now I still have problems with `flash_attn`. I'll try to figure this out tomorrow. If I can't, I commit the test code, but I cannot run the test myself because all tests are ignored (probably due to flash_attn).

xrsrke: @0xkerem i just endabled CI/CD for your PR, so now you could run the unit test in our 8x gpus

0xkerem: Hey! I try to click on each section on github and Google a lot before asking because I don't want to bother you. I couldn't find where I could use this 8x GPUs for my PR. I thought I would be tested automatically after the commit, so I made a small test commit to try this, but in the `checks` tab it says This workflow requires approval from a maintainer. Sorry for your time.

xrsrke: @xrsrke https://github.com/huggingface/nanotron/actions

xrsrke: @0xkerem https://github.com/huggingface/nanotron/actions

xrsrke: ah right, i just approved your cicd

NouamaneTazi:
Hello I see some tests failing because of circular imports. It should be fixed on main now :)

0xkerem:
> Hello I see some tests failing because of circular imports. It should be fixed on main now :)

I saw the notification that the test failed on my phone, I came home a little late but now I'm on the computer. Thanks for reporting the reason for the failed test :)

0xkerem:
https://github.com/huggingface/nanotron/blob/ff3c7746577948743da08c4868aca46cbc0c110b/src/nanotron/parallel/context.py#L128-L130

@NouamaneTazi Hi! Could there be a problem with these lines? The function is defined to return a 3-element tuple, but this code appears to return 4 elements. It also says without expert_parallel_rank in the comment line, but there is nothing to fulfill this condition. Or is there something I missed?

xrsrke (changes requested):
The tests are failed https://github.com/huggingface/nanotron/actions/runs/8189938036/job/22499974452?pr=94

0xkerem:
@xrsrke I think I finally passed the tests I added. There is currently an error left, which is from the main branch and not caused by me. Probably I solved it, but I removed the line I added in my last commit, thinking it might not be needed. I'll re-add that line and try one more time.

src/nanotron/parallel/context.py
```diff
@@ -127,11 +127,35 @@ def set_device(self):
 
     def get_local_ranks(self, world_rank: int) -> Tuple[int, int, int]:
```
xrsrke, line 128: Maybe add a type hint for which parallelism each rank corresponds to. For example [Tuple[Annotated[int, "pp rank"]], ...]

src/nanotron/trainer.py
```diff
@@ -206,9 +206,9 @@ def __init__(
             self.consumed_train_samples = 0
 
         # Setup tensorboard write and log writers on output rank
-        self.logger_ranks = self.parallel_context.world_rank_matrix[
+        self.logger_ranks = self.parallel_context.get_global_rank(
```
xrsrke, line 209: It would be cool if we could get a global rank like get_global_rank(pp_rank=..., tp_rank=...)

tests/test_distributed.py
```diff
@@ -25,6 +25,12 @@ def _test_init_parallel_context(parallel_context: ParallelContext):
     assert isinstance(parallel_context.world_rank_matrix, np.ndarray)
     assert isinstance(parallel_context.world_ranks_to_pg, dict)
 
+    local_rank = tuple(i.item() for i in np.where(parallel_context.world_rank_matrix == world_rank))
+    global_rank = parallel_context.get_global_rank(*local_rank)
+    assert isinstance(global_rank, np.int64), f"The type of global_rank is {type(global_rank)}"
+
+    assert global_rank == parallel_context.world_rank_matrix[local_rank]
```
xrsrke, line 32: global_rank should equal to `dist.get_rank()`

xrsrke (changes requested):
Nice. Thanks for the contribution. Some minor changes

tests/test_distributed.py
0xkerem: @xrsrke Thank you for the feedback! I hope I understood what was asked correctly and implemented the new change request :)

src/nanotron/trainer.py
```diff
@@ -206,9 +206,12 @@ def __init__(
...
         # Setup tensorboard write and log writers on output rank
-        self.logger_ranks = self.parallel_context.world_rank_matrix[
-            0, self.unwrapped_model.output_pp_rank, 0, 0
-        ].flatten()
+        self.logger_ranks = self.parallel_context.get_global_rank(
+            ep_rank=0,
+            pp_rank=self.unwrapped_model.output_pp_rank,
+            dp_rank=0,
+            tp_rank=0
+        ).flatten()
```
NouamaneTazi, line 214: Do we need the flatten()? Can we remove it too?

NouamaneTazi (changes requested):
great job! left one last small comment before merging!!

src/nanotron/trainer.py
0xkerem: I am not sure about that, because the return type of `get_global_rank` is `np.int64`. We use `.flatten()` here is that it converts the return type from a scalar `np.int64` into a `numpy.ndarray`. This conversion is crucial because, in other parts of the code, `self.logger_ranks` is expected to be an array.

Without using `.flatten()`, we could encounter an `IndexError` and a `TypeError` related to trying to iterate over or index into what is essentially a scalar (`numpy.int64`) value. This would require modifications in other parts of the code where `self.logger_ranks` is treated as an array.

https://github.com/huggingface/nanotron/blob/07caff1940609f09096cb4b0fd555441799ea252/src/nanotron/trainer.py#L235-L241

https://github.com/huggingface/nanotron/blob/07caff1940609f09096cb4b0fd555441799ea252/src/nanotron/trainer.py#L431-L432

and other 2 similar occurrences, we can see that `self.logger_ranks` is indeed used as an array.

NouamaneTazi: You're right! Thanks for the details explanation :)

NouamaneTazi:
Amazing job! Thanks for the great PR and your reactivity!! ❤️

0xkerem:
> Amazing job! Thanks for the great PR and your reactivity!! ❤️

You're welcome, and I thank you too! This process and your feedback helped me a lot. :)

PR #95, opened 2024-03-04
[DoReMi] Small refactors

xrsrke:


src/nanotron/trainer.py
```diff
@@ -482,9 +482,13 @@ def train_step_logs(
                     ]
                 )
 
-            if wandb is not None:
+            # NOTE: only one rank writes to wandb
+            if dist.get_rank(self.parallel_context.world_pg) == 0 and wandb is not None:
```
NouamaneTazi, line 486: self.logger_ranks[0] could be better?

NouamaneTazi (approved):
LGTM 🚀 Left a small comment to fix before merging

PR #100, opened 2024-03-11
Fix some bugs

jordane95:
* Now dp should be the third dimension of parallel ranks
* One missing argument

### DIFF
--- a/src/nanotron/serialize/weights.py
+++ b/src/nanotron/serialize/weights.py
@@ -85,6 +85,7 @@ def save_weights(model: nn.Module, parallel_context: ParallelContext, root_folde
 
             else:
                 exp_tp_pp_rank_and_size = None
+                is_expert_sharded = False
 
             path = get_path(
                 base_name,
--- a/src/nanotron/parallel/context.py
+++ b/src/nanotron/parallel/context.py
@@ -126,7 +126,6 @@ class ParallelContext:
         torch.cuda.set_device(torch.cuda.device(device_id))
 
     def get_local_ranks(self, world_rank: int) -> Tuple[int, int, int]:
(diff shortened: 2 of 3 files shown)


src/nanotron/serialize/weights.py
```diff
@@ -85,6 +85,7 @@ def save_weights(model: nn.Module, parallel_context: ParallelContext, root_folde
 
             else:
                 exp_tp_pp_rank_and_size = None
+                is_expert_sharded = False
```
xrsrke, line 88: why do we put is_expert_sharded = False? would be cool if we put some notes here

xrsrke (changes requested):

xrsrke (changes requested):
> Now dp should be the third dimension of parallel ranks
Could you please elaborate on this?

> One missing argument
Where is it 🤔

jordane95:
> > Now dp should be the third dimension of parallel ranks
> 
> Could you please elaborate on this?
> 
> > One missing argument
> 
> Where is it 🤔

1. It is quite straightforward by looking at the definition of rank matrix
https://github.com/huggingface/nanotron/blob/ff3c7746577948743da08c4868aca46cbc0c110b/src/nanotron/parallel/context.py#L69-L76

2. Missing `is_expert_sharded` variable would cause referenced before defined error

https://github.com/huggingface/nanotron/blob/ff3c7746577948743da08c4868aca46cbc0c110b/src/nanotron/serialize/weights.py#L72-L95

NouamaneTazi (approved):
LGTM! it's part of my series of expert parallelism fixes. Some fixes made it in a previous PR. Thanks for noticing!

Issue #45, opened 2024-01-26
[Feature Request] Support Data Streaming for faster training of large models


chagri:
Training large models like 10s or 100s of billions of parameter models require several optimizations. One of the key aspect is tokenize data in advance and then stream during training. Something similar to what MosaicML does with their Streaming and Composer. 

Request to either connect with MosaicML Streaming (open source) or connect with HF data streaming for tokenize in advance and then use it fur training LLMs. This minimizes run time GPU training time and potential failures.

NouamaneTazi:
Thank you @chagri for the detailed issue! But I don't see how would tokenizing data be the bottleneck while training?

I agree though that it'd to great to support Data Streaming from cloud to avoid storing huge datasets locally. Should be supported soon :)

haeggee:
Hey! I'm jumping on this issue because I had a similar question :) 
In short: data streaming from local files is not supported either, right? Essentially, the dataset is tokenized once [here](https://github.com/huggingface/nanotron/blob/9f117eeb1e8309c45e2b1820069292ed5d7c4647/src/nanotron/dataloader.py#L312) and stored in cache files. In certain cases, it might be quite wasteful to both store the raw dataset + the tokenized version -- do you have such local streaming (with tokenization on the fly) on your roadmap?

PR #102, opened 2024-03-13
Adding memmap input data pipelines

TJ-Solergibert:
# What does this PR do?

Solves #45 

The current version of Nanotron's input pipelines is based on Hugging Face Datasets and relies on `clm_preprocess`, which tokenises and preprocesses the entire dataset at the beginning of the training (linked to the `sequence_length`, making it even more difficult to reuse the result across different experiments).

I have developed new data input pipelines based on those included in Megatron. Since I didn't want Nanotron to lose its essence, I removed many functionalities that we don't need (such as those related to BERT models pretraining). What I mainly modified is the `torch.utils.data.Dataset`, and we continue to work with the same Sampler*, Collator* and DataLoader (* I had to modify them slightly), so it doesn't alter the behavior of other modules like the PipelineEngine at all. It also allows us to continue using the previous pipeline based on Hugging Face Datasets, since I added the script `run_train_nanoset.py` to launch the training with the new pipeline.

Relevant details:
- The new input pipelines work with the same files as Megatron. In this PR, I include references to this project to carry out the data preprocessing, although we could include them here as well. The scripts could be simplified, and necessary dependencies could be added.
- I included a new configuration in the .yaml file called [`NanosetDatasetsArgs`](https://github.com/TJ-Solergibert/nanotron/blob/e19dd44cc579e7e94ade44d0c2b4409c283ad2de/src/nanotron/config/config.py#L99), which can replace `PretrainDatasetsArgs. You only need to specify the path to the dataset (generated by Megatron's preprocess_data.py, without the extension as they specify) and the distribution of the dataset samples for each of the splits (train, valid, and test) so that it sums up to 1.
- The [`Nanoset`](https://github.com/TJ-Solergibert/nanotron/blob/e19dd44cc579e7e94ade44d0c2b4409c283ad2de/src/nanotron/data/nanoset.py#L20) will be the new dataset format. It is a lighter version of [`GPTDataset`](https://github.com/NVIDIA/Megatron-LM/blob/89574689447d694bb19dd86fc8a6153b4467ba9d/megatron/core/datasets/gpt_dataset.py#L104C7-L104C17) and [`MegatronDataset`](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/datasets/megatron_dataset.py) from Megatron.
- To build the datasets, we will use the [`NanosetBuilder`](https://github.com/TJ-Solergibert/nanotron/blob/e19dd44cc579e7e94ade44d0c2b4409c283ad2de/src/nanotron/data/dataset_builder.py#L17), which, based on a [`NanosetConfig`](https://github.com/TJ-Solergibert/nanotron/blob/e19dd44cc579e7e94ade44d0c2b4409c283ad2de/src/nanotron/data/nanoset_configs.py#L8) (Contains `NanosetDatasetsArgs` + other details), will build a `Nanoset` for each split. In this first version, we only support one dataset file, but I will include the possibility of using multiple files (BlendedNanoset), hence preserving the `NanosetBuilder`.
- Each `Nanoset` contains an [`MMapIndexedDataset`](https://github.com/TJ-Solergibert/nanotron/blob/e19dd44cc579e7e94ade44d0c2b4409c283ad2de/src/nanotron/data/indexed_dataset.py#L335). This object is found in `indexed_dataset.py` and comes from fairseq. Megatron also includes it as such.
- I have added a [page to the documentation](https://github.com/TJ-Solergibert/nanotron/blob/nanoset/docs/nanoset.md) with more details about the preprocessing to create the datasets and the internal functioning similar to what is included in Megatron.

I think maybe we should centralize the input data pipelines and perhaps move the `dataloader.py` file to another location. I also propose moving several functions from this file with the comments _# Question:_.

To use the Nanoset datasets, you need to specify the `data_path` and `split` fields in `config.data.dataset` in the .yaml file and use the script `run_train_nanoset.py` in the same way as `run_train.py`.

```yaml
data:
  dataset:
    data_path: /mloscratch/homes/solergib/s-ai/nanotron/datasets/llama2/europarl-gpt-llama2_text_document
    split: 949,50,1
  num_loading_workers: 0
  seed: 1234
```

I've published the [wandb logs](https://wandb.ai/tj-solergibert/nanoset_dataloader/overview) of the different tests I have carried out, comparing the HF Datasets and the new Nanoset datasets with 1 and 4 GPUs and resuming training from a checkpoint. 

This is a first version, I am open to all suggestions you can think of! I have named it Nanoset, but as I said, you are free to change it!

Toni


NouamaneTazi:
Amazing job 🚀 Feel free to request reviews once you feel the PR is ready!

TJ-Solergibert:
Hello! Everything is ready now. I've added the following:
- [`Nanoset`](https://github.com/huggingface/nanotron/blob/5984cf4a3e73b2f6872e847a39bbb7390d4f624c/src/nanotron/data/nanoset.py#L20): A new dataset inspired by Megatron.
- [`BlendedNanoset`](https://github.com/huggingface/nanotron/blob/5984cf4a3e73b2f6872e847a39bbb7390d4f624c/src/nanotron/data/blended_nanoset.py#L20): Similar to the previous one, but this one is for mixing different datasets. It allows specifying the weight of each dataset.
- [`MMapIndexedDataset`](https://github.com/huggingface/nanotron/blob/5984cf4a3e73b2f6872e847a39bbb7390d4f624c/src/nanotron/data/indexed_dataset.py#L306): Both `Nanoset` and `BlendedNanoset` contain `MMapIndexedDataset`, where the data is actually stored ([__check__](https://github.com/huggingface/nanotron/blob/5984cf4a3e73b2f6872e847a39bbb7390d4f624c/src/nanotron/data/nanoset.py#L142), as it queries the data using the `get` method of `MMapIndexedDataset`).
- [`NanosetBuilder`](https://github.com/huggingface/nanotron/blob/5984cf4a3e73b2f6872e847a39bbb7390d4f624c/src/nanotron/data/dataset_builder.py#L19): Responsible for creating the train, valid, and test datasets whether they are Nanoset or BlendedNanoset type.
- New [`NanosetDatasetsArgs`](https://github.com/huggingface/nanotron/blob/5984cf4a3e73b2f6872e847a39bbb7390d4f624c/src/nanotron/config/config.py#L99) to configure the new datasets. I've also added `val_steps` to [`TokensArgs`](https://github.com/huggingface/nanotron/blob/5984cf4a3e73b2f6872e847a39bbb7390d4f624c/src/nanotron/config/config.py#L226) to generate valid and train datasets.
- Tools for: 
  - Preprocessing the data and generating Megatron-style datasets ([`tools/preprocess_data.py`](https://github.com/huggingface/nanotron/blob/5984cf4a3e73b2f6872e847a39bbb7390d4f624c/tools/preprocess_data.py)). Inspired by [`Megatron-LM/tools/preprocess_data.py`](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/preprocess_data.py), I have developed a simpler version, getting rid of several features we don't need, such as split sentences or multimodal datasets. I've also enforced the tokenizer to be built with `AutoTokenizer.from_pretrained`. I've kept the input format of Megatron, which is based on a .json file containing the samples to make it easier to transition to Nanotron. 
  - A small script to transform datasets from the Hugging Face Hub (or local) to the .json format needed by `tools/preprocess_data.py` ([`tools/hf_datasets_to_json.py`](https://github.com/huggingface/nanotron/blob/5984cf4a3e73b2f6872e847a39bbb7390d4f624c/tools/hf_datasets_to_json.py)). 
  - A tool to merge already preprocessed datasets into a single file ([`tools/merge_datasets.py`](https://github.com/huggingface/nanotron/blob/5984cf4a3e73b2f6872e847a39bbb7390d4f624c/tools/merge_datasets.py)).

In the [`docs`](https://github.com/huggingface/nanotron/blob/cde43c72ca03479ca34555ab27ba970f67040951/docs/nanoset.md), I've added instructions for using the new datasets. As you can see, it maintains compatibility with the original input data pipelines and with the rest of the project. Basically, the only thing I've modified is the `torch.utils.data.Dataset`, and we continue to work with practically the same collators and samplers and the same dataloader.

Some small comments and doubts that have arisen:
- In [`src/nanotron/dataloader.py`](https://github.com/huggingface/nanotron/blob/5984cf4a3e73b2f6872e847a39bbb7390d4f624c/src/nanotron/dataloader.py#L244), there is the `SkipBatchSampler`, `EmptyInfiniteDataset`, and `get_dataloader_worker_init` that I use in [`src/nanotron/data/dataloader_builder.py`](https://github.com/huggingface/nanotron/blob/5984cf4a3e73b2f6872e847a39bbb7390d4f624c/src/nanotron/data/dataloader_builder.py). Perhaps we could move them to `nanotron/data/dataloader_builder.py` to centralize this section of the project a bit.
- In the `log_rank` function of [`src/nanotron/logging.py`](https://github.com/huggingface/nanotron/blob/5984cf4a3e73b2f6872e847a39bbb7390d4f624c/src/nanotron/logging.py#L211), I've added a try-except. I did this to allow running preprocessing scripts without initializing a distributed process group. We can leave it like this or initialize the distributed process group within the preprocessing files.
- In [`tools/preprocess_data.py`](https://github.com/huggingface/nanotron/blob/5984cf4a3e73b2f6872e847a39bbb7390d4f624c/tools/preprocess_data.py), I've added a small trick to preprocess the data without installing Nanotron. Let me know if you think it's interesting or if we should get rid of it.
- To preprocess the data with the Hugging Face tokenizers, we would need to add transformers as a dependency of the project.

In short: `NanosetBuilder` is responsible for building the train, valid, and test datasets. Depending on whether we specify a blend of multiple datasets or not, it will create either a `BlendNanoset` or a `Nanoset` for each split. Each `Nanoset` contains an `MMapIndexedDataset`, which is the dataset that contains and reads the bytes with the tokens from the files generated by `preprocess_data.py`, while the `Nanoset` itself contains "positions" to read from the `MMapIndexedDataset` for each sample. The `BlendNanoset` contains one `Nanoset` with the samples to extract from the `MMapIndexedDataset` for each specified path, complying with the specified weights.

Looking forward hearing your feedback!

Toni

cc @NouamaneTazi

TJ-Solergibert:
Hello! 

I have simplified the dataset builder, the `__getitem__` method from Nanoset, and fixed a minor bug with the parser that couldn't properly identify the `NanosetDatasetsArgs`.

Toni

xrsrke:
@TJ-Solergibert Hi. Thanks for the fantastic PR. Would be cool if you can add some unit tests for `build_nanoset_dataloader`, and `NanosetBuilder(...).build()`!

TJ-Solergibert:
Hello! I've just added the tests. 

I tried to stick to the design of the other tests in the repository, but it was also the first time I was developing one 😅. Finally, I've designed a script that tests everything I've included: It starts by creating a `.json` file like the ones the `preprocess_data.py` script expects and processes it to generate the `.idx` and `.bin` files containing the tokens. Then, we verify that we can create each type of `Nanosets` (`Nanoset` and `BlendedNanoset`) and create the `Dataloader`. Finally, we check that the content of the batches in each and every process is appropriate.

In this verification ([`assert_batch_dataloader`](https://github.com/huggingface/nanotron/blob/d2238359cbc55650d1f2e0e1dfc01c71c36b26e7/tests/helpers/data.py#L54)), we ensure that the content of each element of the batch (`input_ids`, `input_mask`, `label_ids` & `label_mask`) is exactly the same across processes within the same _tensor parallel group_, and also that the class of each element (distinguishing between tensors with ids, tensors with masks, and `TensorPointer`) is exactly the same across processes within the same _data parallel group_.

I have locally run the pytest on a cluster with up to 8 GPUs, and they have been satisfactory (It hurt a little to waste the GPUs, but for the development of the test, I did it by applying a small patch to be able to start the dist group with "gloo"). I look forward to your comments!

Toni

TJ-Solergibert:
Hello! 

I think it's a good time for a review. I've further cleaned up the project and made sure to thoroughly document the operation of the Nanosets. Below, I provide all the organized information.

# Nanoset
[`Nanosets`](https://github.com/TJ-Solergibert/nanotron/blob/nanoset/src/nanotron/data/nanoset.py) are a new type of dataset for Nanotron inspired by those of Megatron. They maintain their performance by dispensing with unnecessary features and (trying to) perserving the essence of Nanotron. In essence, it's just a `torch.utils.data.Dataset`, so we maintain compatibility with the rest of the project by slightly modifying some aspects related to data loading, such as the collator or the DistributedSampler. Inside, each Nanoset has an [`MMapIndexedDataset`](https://github.com/TJ-Solergibert/nanotron/blob/ce4e7432731fd375b9238f0d54e17dc79e79390b/src/nanotron/data/indexed_dataset.py#L302) from which we extract the tokens to build the samples. The main task of the `Nanoset` is to control the logic for constructing the samples from the tokens of the `MMapIndexedDataset`.
# BlendedNanoset
The [`BlendedNanoset`](https://github.com/TJ-Solergibert/nanotron/blob/nanoset/src/nanotron/data/blended_nanoset.py) is used to create a mixture of `Nanosets` by specifying the weights for each of them. Essentially, the `BlendedNanoset` is only responsible of ensuring that the dataset indices comply with the specified blend, as the samples are extracted from the `Nanosets`.
# NanosetDatasetArgs
To use the Nanosets, I have added a new configuration to the config .yaml file ([`NanosetDatasetsArgs`](https://github.com/TJ-Solergibert/nanotron/blob/ce4e7432731fd375b9238f0d54e17dc79e79390b/src/nanotron/config/config.py#L99)). The user only needs to enter the dataset(s) they want to use, how they want to divide the dataset into train, valid, and test partitions, and optionally a directory to store `Nanoset` metadata for reusing the same configuration across different runs.

# Tools
Added tools for: 
- Preprocessing the data and generating Megatron-style datasets ([`tools/preprocess_data.py`](https://github.com/TJ-Solergibert/nanotron/blob/nanoset/tools/preprocess_data.py)). Inspired by [`Megatron-LM/tools/preprocess_data.py`](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/preprocess_data.py), I have developed a simpler version, getting rid of several features we don't need, such as split sentences or multimodal datasets. I've also enforced the tokenizer to be built with `AutoTokenizer.from_pretrained`. I've kept the input format of Megatron, which is based on a .json file containing the samples to make it easier to transition to Nanotron. 
- A small script to transform datasets from the Hugging Face Hub (or local) to the .json format needed by `tools/preprocess_data.py` ([`tools/hf_datasets_to_json.py`](https://github.com/TJ-Solergibert/nanotron/blob/nanoset/tools/hf_datasets_to_json.py)). 
- A tool to merge already preprocessed datasets into a single file ([`tools/merge_datasets.py`](https://github.com/TJ-Solergibert/nanotron/blob/nanoset/tools/merge_datasets.py)).

# Docs
In the [`docs`](https://github.com/TJ-Solergibert/nanotron/blob/nanoset/docs/nanoset.md), I've added instructions for preprocessing the data, using the Nanosets and how do the work under the hood. **I strongly recommend taking a look at the examples of how the samples are constructed to understand the operation of the Nanosets.**

# Short summary

In short: [`NanosetBuilder`](https://github.com/TJ-Solergibert/nanotron/blob/nanoset/src/nanotron/data/dataset_builder.py) is responsible for building the train, valid, and test datasets. Depending on whether we specify a blend of multiple datasets or not, it will create either a `BlendedNanoset` or a `Nanoset` for each split. Each `Nanoset` contains an `MMapIndexedDataset`, which is the dataset that contains and reads the bytes with the tokens from the files generated by `preprocess_data.py`, while the `Nanoset` itself contains "positions" to read from the `MMapIndexedDataset` for each sample. The `BlendedNanoset` contains one `Nanoset` with the samples to extract from the `MMapIndexedDataset` for each specified path, complying with the specified weights.

# Tests
I tried to stick to the design of the other tests in the repository, but it was also the first time I was developing one 😅. Finally, I've designed a script that tests everything I've included: It starts by creating a `.json` file like the ones the `preprocess_data.py` script expects and processes it to generate the `.idx` and `.bin` files containing the tokens. Then, we verify that we can create each type of `Nanosets` (`Nanoset` and `BlendedNanoset`) and create the `Dataloader`. Finally, we check that the content of the batches in each and every process is appropriate.

In this verification ([`assert_batch_dataloader`](https://github.com/TJ-Solergibert/nanotron/blob/ce4e7432731fd375b9238f0d54e17dc79e79390b/tests/helpers/data.py#L50)), we ensure that the content of each element of the batch (`input_ids`, `input_mask`, `label_ids` & `label_mask`) is exactly the same across processes within the same _tensor parallel group_, and also that the class of each element (distinguishing between tensors with ids, tensors with masks, and `TensorPointer`) is exactly the same across processes within the same _data parallel group_. Also, in the case of `BlendedNanoset`, we verify that it is composed of more than one `Nanoset` and that each `Nanoset` has enough samples to satisfy the `BlendedNanoset`.

# Python vs CPP
In the last commit, I replaced the functions we were using from CPP with Python functions. As expected and as seen in [this notebook](https://drive.google.com/file/d/1iUWeeIVBMJKRypjoV5lHAnHjM1oOWt6B/view?usp=sharing), the CPP functions are incredibly faster than the Python ones, even causing the training start to be blocked for minutes. In the case of the function used to create the indices of the `BlendedNanoset` ([`build_blending_indices`](https://github.com/TJ-Solergibert/nanotron/blob/ce4e7432731fd375b9238f0d54e17dc79e79390b/src/nanotron/data/utils.py#L72)), in Python, we would spend 44 seconds to create a dataset of 1e7 samples and over 7 minutes for 1e8 samples, while in CPP, we would only wait for 0.1 and 1 second, respectively. Regarding the function used in each `Nanoset` to build samples of sequence length from the documents ([`build_sample_idx`](https://github.com/TJ-Solergibert/nanotron/blob/ce4e7432731fd375b9238f0d54e17dc79e79390b/src/nanotron/data/utils.py#L102)), we spend 35 seconds if we have 1e7 documents (each entry in the json file is considered a document), while the same function in CPP would take 0.5 seconds. I believe it is incredibly worthwhile to use the ones developed in CPP.

# Other comments
- In [`src/nanotron/dataloader.py`](https://github.com/TJ-Solergibert/nanotron/blob/ce4e7432731fd375b9238f0d54e17dc79e79390b/src/nanotron/dataloader.py#L244), there is the `SkipBatchSampler`, `EmptyInfiniteDataset`, and `get_dataloader_worker_init` that I use in [`src/nanotron/data/dataloader_builder.py`](https://github.com/TJ-Solergibert/nanotron/blob/ce4e7432731fd375b9238f0d54e17dc79e79390b/src/nanotron/data/dataloader_builder.py#L11). Perhaps we could move them to `nanotron/data/dataloader_builder.py` to centralize this section of the project a bit.
- In the `log_rank` function of [`src/nanotron/logging.py`](https://github.com/TJ-Solergibert/nanotron/blob/ce4e7432731fd375b9238f0d54e17dc79e79390b/src/nanotron/logging.py#L211), I've added a try-except. I did this to allow running the preprocessing scripts without initializing a distributed process group. We can leave it like this or initialize the distributed process group within the preprocessing files.
- In [`tools/preprocess_data.py`](https://github.com/TJ-Solergibert/nanotron/blob/ce4e7432731fd375b9238f0d54e17dc79e79390b/tools/preprocess_data.py#L9), I've added a small trick to preprocess the data without installing Nanotron. Let me know if you think it's interesting or if we should get rid of it.
- To preprocess the data with the Hugging Face tokenizers, I added `transformers` to the requirements of the project.

That would be all, tell me what you think!

Toni

xrsrke (commented):
very cool.

xrsrke (commented):
Very cool. But should we move this license to somewhere specific to Megatron's dataset? Because most of the Nanotron code isn't from Megatron (this is a mistake... will give more detailed review)

src/nanotron/config/config.py
```diff
@@ -210,6 +221,7 @@ class TokensArgs:
     batch_accumulation_per_replica: int
 
     val_check_interval: Optional[int] = -1
+    val_steps: Optional[int] = 10
```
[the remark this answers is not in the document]
TJ-Solergibert: This is another detail I wanted to discuss. Currently, the `len(train_dataset)` is calculated as the `number of training steps * global batch size`, while for evaluation/test, what you need is to specify how often you want it to be conducted and how many steps you want to perform. From here, you can determine the total number of samples you need to create each dataset.

An example will make this clearer: Let's start with a preprocessed dataset that has 100 documents, and with a split of `8,1,1`, 80 documents will go to train, 10 to valid, and 10 to test. With these 80 documents for training, let's say you extract 50 training samples. If you have a global batch size of 8 and 10 training steps, you will need to generate 80 samples, meaning you will train for 1.4 epochs. 

However, for validation, if you have an `val_check_interval` of 2, during 10 training steps, you will conduct the evaluation 6 times. In these 6 evaluations, you will perform `val_steps`, which if it's 10, you will need your valid dataset to have `6 * 10 * global batch size` samples. Considering that perhaps from the 10 documents we used in the example, you only extract 5 samples and the global batch size is 8, in each of the 6 validation checks, you consume 80 samples, which means you will consume those 5 samples 12 times.

I don't see the point in this method, as I believe it would be better for both validation and test to consume the entire dataset only once. I will modify it.

src/nanotron/data/dataloader_builder.py
```diff
@@ -0,0 +1,166 @@
...
+                f" {self.sequence_length}."
+            )
+
+        # Cast np.array to torch.Tensor
+        result = {k: v if isinstance(v, TensorPointer) else torch.from_numpy(v) for k, v in result.items()}
+        return result
+
+
+def get_sampler(
+    dl_ranks_size: int,
```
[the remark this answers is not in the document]
TJ-Solergibert: I copied from [here](https://github.com/huggingface/nanotron/blob/f7a021b8fa84dbebfbb575c4db7afd771fab0f66/src/nanotron/dataloader.py#L401).

src/nanotron/data/dataset_builder.py
```diff
@@ -0,0 +1,214 @@
...
+        # Single Nanoset
+        if isinstance(data_path, str):
+            return self.build_nanoset_dataset_splits(data_path, split, self.config.split_num_samples)
+
+        # Blended Nanoset
+        prefix_per_dataset = list(data_path.keys())
+        weight_per_dataset = normalize(list(data_path.values()))
+
+        # NOTE: Use 0.5% target margin to ensure that that we do not exhaust the indices of any dataset in the Blend
+        sizes_per_dataset = [
```
xrsrke, line 58: We should add some logging to track the number of samples per dataset based on the weights and the total number of samples after weighting?

src/nanotron/data/dataset_builder.py
```diff
@@ -0,0 +1,214 @@
...
+        path_prefix: str,
+        split: List[float],
+        sizes: List[int],
+    ) -> List[Nanoset]:
+        """Build each Nanoset split from a single MMapIndexedDataset
+
+        Args:
+            path_prefix (str): The MMapIndexedDataset .bin and .idx file prefix
+
+            split (List[float]): The dataset split ratios (must sum to 1.00)
```
xrsrke, line 112: use `weights` as the argument name?

xrsrke: and assert it sum up to 1

tests/test_build_nanoset_dataloader.py
```diff
@@ -0,0 +1,130 @@
...
+
+@pytest.mark.parametrize(
+    "tp,dp,pp",
+    [
+        pytest.param(*all_3d_configs)
+        for gpus in range(1, min(available_gpus(), 4) + 1)
+        for all_3d_configs in get_all_3d_configurations(gpus)
+    ],
+)
+@rerun_if_address_is_in_use()
```
xrsrke, line 30: We should add a separate test to verify that, given the same seed, NanoSet is deterministic while also varying sequence_length, the total number of samples, split ratio... (yes, I am aware that currently in nanotron we also don't have these tests for the current dataloader, but from now on, we aim to rectify this by adding as many good tests as possible)

tests/test_build_nanoset_dataloader.py
```diff
@@ -0,0 +1,130 @@
...
+        split_num_samples=split_num_samples,
+        path_to_cache=test_context.get_auto_remove_tmp_dir(),
+    )
+
+    configs = [nanoset_config, blended_nanoset_config]
+    dataset_types = [Nanoset, BlendedNanoset]
+
+    for config, dataset_type in zip(configs, dataset_types):
+        # Create Nanosets
+        train_dataset, valid_dataset, test_dataset = NanosetBuilder(config).build()
```
xrsrke, line 99: We should add a separate test to check the dataset length, sequence length, and non-overlap (ensuring dataset[1] != dataset[2]) in a BlendedNanoSet.

tests/test_build_nanoset_dataloader.py
```diff
@@ -0,0 +1,130 @@
...
+        assert isinstance(train_dataset, dataset_type)
+        if isinstance(train_dataset, BlendedNanoset):
+            assert len(train_dataset.datasets) > 1
+            for idx, dataset in enumerate(train_dataset.datasets):
+                assert len(dataset) > get_max_value_by_group(
+                    train_dataset.dataset_index, train_dataset.dataset_sample_index, idx
+                )
+
+        # Create Dataloaders
+        dataloader = build_nanoset_dataloader(
```
xrsrke, line 110: check non-overlapse across dp ranks like https://github.com/huggingface/nanotron/blob/f7a021b8fa84dbebfbb575c4db7afd771fab0f66/examples/doremi/tests/test_doremi_sampler.py#L76

and not repeating samples like https://github.com/huggingface/nanotron/blob/f7a021b8fa84dbebfbb575c4db7afd771fab0f66/examples/doremi/tests/test_doremi_sampler.py#L319

tests/test_build_nanoset_dataloader.py
```diff
@@ -0,0 +1,130 @@
...
+            input_pp_rank=input_pp_rank,
+            output_pp_rank=output_pp_rank,
+            micro_batch_size=MICRO_BATCH_SIZE,
+            dataloader_num_workers=0,
+            dataloader_drop_last=True,
+        )
+
+        # Check a batch produced by the Dataloader
+        batch = next(iter(dataloader))
+        assert_batch_dataloader(
```
xrsrke, line 123: + check only input_pp_rank and output_pp_rank return the data

tests/test_build_nanoset_dataloader.py
```diff
@@ -0,0 +1,130 @@
...
+    )
+
+
+def _test_build_nanoset_dataloader(
+    parallel_context: ParallelContext, test_context: TestContext, path_to_bin_files: str
+):
+    SEED = 1234
+    random.seed(SEED)
+
+    TRAIN_STEPS = random.randint(1000, 10000)
```
xrsrke, line 56: For easier to detect what cause a test fail later on, maybe we should fix the train_steps? Parameterize it [100, 500] or something

tests/test_build_nanoset_dataloader.py
```diff
@@ -0,0 +1,130 @@
...
+    VAL_CHECK_INTERVAL = TRAIN_STEPS / 4
+    VAL_STEPS = TRAIN_STEPS / 10
+    MICRO_BATCH_SIZE = 2 ** random.randint(1, 4)
```
xrsrke, line 59: same as above

tests/test_build_nanoset_dataloader.py
```diff
@@ -0,0 +1,130 @@
...
+    N_MICRO_BATCHES_PER_BATCH = 2 ** random.randint(1, 4)
```
xrsrke, line 60: same as above

tests/test_build_nanoset_dataloader.py
```diff
@@ -0,0 +1,130 @@
...
+    GLOBAL_BATCH_SIZE = MICRO_BATCH_SIZE * N_MICRO_BATCHES_PER_BATCH * parallel_context.dp_pg.size()
+
+    SEQ_LENGTH = 2 ** random.randint(10, 14)
```
xrsrke, line 63: same as above

src/nanotron/config/config.py
```diff
@@ -95,13 +95,24 @@ def __post_init__(self):
             self.hf_dataset_splits = "train"
 
 
+@dataclass
+class NanosetDatasetsArgs:
+    data_path: Union[str, dict]
+    split: Optional[str] = "949,50,1"
+    path_to_cache: Optional[str] = "_"
```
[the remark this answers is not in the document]
TJ-Solergibert: When the default was `None`, dacite wasn't able to correctly parse the config. The fix I came up with was to set a default dummy value, let dacite parse the config, and then as the dummy value (`_`) is not a path it sets the value to `None`.

src/nanotron/config/config.py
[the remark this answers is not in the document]
TJ-Solergibert: I have updated the logic behind the number of samples for each split. Finally, in `NanosetConfigs`, we introduce the quantity of `train_samples` (computed as before, `train_steps * global_batch_size`. Previously, we introduced `split_num_samples` as a list with the quantity of samples per split calculated as I mentioned in my previous message) and from here, after initialization we will create`self.split_num_samples = [train_split_samples, None, None]`. These `None` values will indicate to the Nanoset constructor that its samples will only be consumed once, that is, we will not concatenate the indices as we do for the train split. This way, the `DataLoader`will serve all samples only once.

For the case of the `BlendedNanoset`, based on the value of `train_split_samples` and with the `weights` of each dataset, we will compute the number of samples to extract from each dataset in the blend. In this case, it should be noted that it is possible that one dataset from the blend is consumed 5 times while another larger one is consumed only once. For the valid and test split, we will consume all the samples reserved for these splits only once.

I have added all to the docs.

LICENSE
```diff
@@ -199,3 +199,63 @@
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
+
+This repository also contains code from Facebook (from their Fairseq project)
```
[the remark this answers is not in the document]
TJ-Solergibert: This is something I highly doubt, because currently we are using:

- From Megatron, since we have transferred the CPP functions to Python, we would not be using anything exactly the same as them. We have simply used them as inspiration, but there are no literal copies.

- What I have not modified to a large extent is `indexed_dataset.py`, originally from Megatron, but they themselves cite Fairseq. Although, as you can see, they also mention that they have rewritten it (The script in Fairseq does indeed resemble it a bit, but it is not a literal copy).

indexed_dataset.py [[NANOTRON](https://github.com/huggingface/nanotron/blob/3dd5f62284d34382789f53f486bd2091d9db2237/src/nanotron/data/indexed_dataset.py)] [[MEGATRON](https://github.com/TJ-Solergibert/Megatron-debug/blob/c5eda947d9728d21b03d77b7db56cb71513d5636/megatron/core/datasets/indexed_dataset.py#L6)] [[FAIRSEQ](https://github.com/facebookresearch/fairseq/blob/bedb259bf34a9fc22073c13a1cee23192fa70ef3/fairseq/data/indexed_dataset.py#L391)]

I moved the LICENSE.

src/nanotron/data/dataloader_builder.py
```diff
@@ -0,0 +1,163 @@
...
+        collate_fn=data_collator,
+        drop_last=dataloader_drop_last,
+        num_workers=dataloader_num_workers,
+        pin_memory=dataloader_pin_memory,
+        worker_init_fn=get_dataloader_worker_init(dp_rank=dp_rank),
+    )
+
+
+@dataclass
+class NanosetDataCollatorForCLM:
```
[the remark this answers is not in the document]
TJ-Solergibert: Done. At the beginning I though that maybe I had to modify something else.

src/nanotron/data/dataloader_builder.py
```diff
@@ -0,0 +1,163 @@
...
+                f"`labels` are incorrectly preprocessed. `labels` length is {result['label_ids'].shape[-1]}, but should be"
+                f" {self.sequence_length}."
+            )
+
+        # Cast np.array to torch.Tensor
+        result = {k: v if isinstance(v, TensorPointer) else torch.from_numpy(v) for k, v in result.items()}
+        return result
+
+
+def get_sampler(
```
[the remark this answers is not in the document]
TJ-Solergibert: Although at the moment we do not use the valid and test splits, I have implemented the creation of the DataLoaders, so the name `_get_train_sampler` did not convince me since I also use it for the valid and test splits. Shall we change the name to `_get_sampler`? `_get_dataloader_sampler`?

src/nanotron/data/dataset_builder.py
TJ-Solergibert: There are two different distributions:
1. The train, valid, and test split. This is the one you are mentioning, which is entered in the split field of the .yaml file. When creating the `NanosetConfig`, we already parse and normalize this string.

2. The weights of each dataset for the `BlendedNanoset`. These are specified in the .yaml file, where the values are the weights. In the `dataset_builder`, we use them to compute the number of samples for each `Nanoset`.

TJ-Solergibert: Should we add the assert considering that we always normalize it?

tests/test_build_nanoset_dataloader.py
TJ-Solergibert: I have parameterized the number of train steps and the sequence length (They are the most relevant parameters). I have set the micro batch size and the number of micro batches per batch to a fixed value.

src/nanotron/logging.py
```diff
@@ -217,12 +217,16 @@ def log_rank(
     **kwargs,
 ):
     """Log only if the current process is the rank specified."""
-    # Use default group is group is not provided
-    if group is None:
-        group = torch_dist.distributed_c10d._get_default_group()
+    # Use default group if group is not provided
+    try:
+        if group is None:
+            group = torch_dist.distributed_c10d._get_default_group()
```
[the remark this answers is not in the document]
TJ-Solergibert: Changed with `if torch.distributed.is_initialized():`. The problem is that in the preprocess script we use the `log_rank` which requires to start the distributed group.

tests/test_build_nanoset_dataloader.py
TJ-Solergibert: Indeed, the three variables you mentioned are the ones that most influence the creation of the `Nanoset`. It is worth noting that when preprocessing the JSON file, what we do is tokenize each document (1 line of the JSON = 1 document) and concatenate all the tokens. That's why...

- When varying the sequence length, the samples you construct in the [`sample_index`](https://github.com/TJ-Solergibert/nanotron/blob/16f44d1b8177b38a7ae3a2322a912562dbdc244a/src/nanotron/data/nanoset.py#L212) are different because what you actually do is take sequence length tokens + 1 consecutively from the `document_index` documents.
- When varying the number of samples, you increase or decrease the [number of times](https://github.com/TJ-Solergibert/nanotron/blob/16f44d1b8177b38a7ae3a2322a912562dbdc244a/src/nanotron/data/nanoset.py#L234) you concatenate `shuffle_index` so that it is always greater than the number of samples and does not end the dataset.
- When varying the split, you change the number of documents used to create the Nanoset, so the result of [shuffling the `document_index`](https://github.com/TJ-Solergibert/nanotron/blob/16f44d1b8177b38a7ae3a2322a912562dbdc244a/src/nanotron/data/nanoset.py#L196) and the quantity of samples (and the samples itself) in `sample_index` will be different.

It is important to remember that EVERYTHING random, both the shuffling of the [`document_index](https://github.com/TJ-Solergibert/nanotron/blob/16f44d1b8177b38a7ae3a2322a912562dbdc244a/src/nanotron/data/nanoset.py#L196) and the [`shuffle_index`](https://github.com/TJ-Solergibert/nanotron/blob/16f44d1b8177b38a7ae3a2322a912562dbdc244a/src/nanotron/data/nanoset.py#L231), is done with a `numpy.random.RandomState` and the seed.

tests/test_build_nanoset_dataloader.py
TJ-Solergibert: The `BlendedNanoset` essentially determines which samples to extract from each Nanoset based on the requested [number of samples](https://github.com/TJ-Solergibert/nanotron/blob/16f44d1b8177b38a7ae3a2322a912562dbdc244a/src/nanotron/data/blended_nanoset.py#L27) for the `BlendedNanoset`, the [weights](https://github.com/TJ-Solergibert/nanotron/blob/16f44d1b8177b38a7ae3a2322a912562dbdc244a/src/nanotron/data/blended_nanoset.py#L25) of each `Nanoset`, and their respective lengths in [`build_blending_indices`](https://github.com/TJ-Solergibert/nanotron/blob/16f44d1b8177b38a7ae3a2322a912562dbdc244a/src/nanotron/data/blended_nanoset.py#L129). As can be seen [here](https://github.com/TJ-Solergibert/nanotron/blob/16f44d1b8177b38a7ae3a2322a912562dbdc244a/src/nanotron/data/blended_nanoset.py#L199), when establishing the index of the sample to extract from a specific `Nanoset`, the modulo operation with the length of the `Nanoset` is applied, so we NEVER attempt to access an index greater than the length of the `Nanoset`.

The number of samples required for the `BlendedNanoset` is computed during the initialization of [`NanosetConfig`](https://github.com/TJ-Solergibert/nanotron/blob/16f44d1b8177b38a7ae3a2322a912562dbdc244a/src/nanotron/data/nanoset_configs.py#L54), and the number of samples each` Nanoset` in the blend must have is computed [here](https://github.com/TJ-Solergibert/nanotron/blob/16f44d1b8177b38a7ae3a2322a912562dbdc244a/src/nanotron/data/dataset_builder.py#L58). Note that you can create a `Nanoset` from a file containing 10 documents, resulting in only 5 samples, and request it to have 200; in this case, it will only increase the number of times the indices are [concatenated](https://github.com/TJ-Solergibert/nanotron/blob/16f44d1b8177b38a7ae3a2322a912562dbdc244a/src/nanotron/data/nanoset.py#L236) (You will perform 40 epochs).

tests/test_build_nanoset_dataloader.py
TJ-Solergibert: I don't quite understand what I need to check, as we use PyTorch's [DistributedSampler](https://pytorch.org/docs/stable/data.html#torch.utils.data.distributed.DistributedSampler). In all processes, we create the same dataset with all the indices, and the DistributedSampler is responsible for extracting the correct indices in each process (based on `dp_rank` and `dp_size`)

docs/nanoset.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Working with Nanosets
+
+To work with Nanosets, we need to configure 3 arguments:
+1. `split`: The distribution we want to divide our dataset into among train, valid, and test split.
+2. `path_to_cache`: Directory used to store certain metadata of Nanosets to reuse them between different runs.
+3. `data_path`: This argument specifies the file/files that will compose the Nanoset. There are 2 ways to specify it:
+   1. If we only indicate _one_ path (path to the files generated by preprocess_data.py WITHOUT the extension), we will create a `Nanoset`.
+    ```yaml
+    data:
```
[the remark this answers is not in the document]
TJ-Solergibert: I've just seen the changes. For now, I've adapted the code to keep it running. I still need to enable the users create more dataloaders according to the training stages. The main issue I see is that [here](https://github.com/huggingface/nanotron/blob/531f802a215a605a9d75f770ea83ba9da600b52b/src/nanotron/trainer.py#L315) it accesses `hf_dataset_or_datasets` and `NanosetDatasetArgs` doesn't contain this field (`data_path` instead).

Thank you for the feedback!

Toni

tests/test_build_nanoset_dataloader.py
xrsrke: > When varying the sequence length, the samples you construct in the [sample_index](https://github.com/TJ-Solergibert/nanotron/blob/16f44d1b8177b38a7ae3a2322a912562dbdc244a/src/nanotron/data/nanoset.py#L212) are different because what you actually do is take sequence length tokens + 1 consecutively from the document_index documents.

Yep, I mean, given the same set of sequence length, split ratio, the total number of samples, and also the same seed, then we expect that NanoSet is deterministic

src/nanotron/data/dataset_builder.py
xrsrke: > Should we add the assert considering that we always normalize it?
Yes, it would be nice from an end-user perspective, so they don't have unexpected normalization (like the weights they get should be the weights that they enter)

src/nanotron/data/dataloader_builder.py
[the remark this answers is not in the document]
xrsrke: > Shall we change the name to _get_sampler?
`_get_sampler` is nice

src/nanotron/config/config.py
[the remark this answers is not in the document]
xrsrke: Oh i see. We should put a comment here for later reference!

tests/test_build_nanoset_dataloader.py
xrsrke: > The BlendedNanoset essentially determines which samples to extract from each Nanoset based on the requested number of samples for the BlendedNanoset
Oops. Thanks for clarifying.. We should add the suggested test above for `Nanoset`

tests/test_build_nanoset_dataloader.py
xrsrke: Yeah right 😁

src/nanotron/config/config.py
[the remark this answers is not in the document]
TJ-Solergibert: Done!

src/nanotron/data/dataloader_builder.py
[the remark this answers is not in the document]
TJ-Solergibert: Done!

src/nanotron/data/dataset_builder.py
TJ-Solergibert: [Done!](https://github.com/TJ-Solergibert/nanotron/blob/531f802a215a605a9d75f770ea83ba9da600b52b/src/nanotron/data/dataset_builder.py#L93)

src/nanotron/data/dataset_builder.py
TJ-Solergibert: [Done!](https://github.com/TJ-Solergibert/nanotron/blob/f163c1d6574b5f8f1d7e9d51caa9f006ec35660a/src/nanotron/data/dataset_builder.py#L115)

tests/test_build_nanoset_dataloader.py
TJ-Solergibert: Okey! [Done](https://github.com/TJ-Solergibert/nanotron/blob/f163c1d6574b5f8f1d7e9d51caa9f006ec35660a/tests/helpers/data.py#L119)!

We check all the important features and we extract a sample from the dataset (sample with index 23).

tests/test_build_nanoset_dataloader.py
TJ-Solergibert: [Done!](https://github.com/TJ-Solergibert/nanotron/blob/f163c1d6574b5f8f1d7e9d51caa9f006ec35660a/tests/helpers/data.py#L70)

docs/nanoset.md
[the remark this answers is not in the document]
TJ-Solergibert: I've implemented the compatibility with the training stages. I've also found a bug which I solved in this [commit](https://github.com/huggingface/nanotron/pull/102/commits/05be999d312855069b9cf1723d5c1c68ac5a431e). I've uploaded some [logs](https://pastebin.com/SzuFyfx5) of a run in which the training stages works smoothly.

TJ-Solergibert:
Hello!

I have finally polished the two remaining fronts: `MMapIndexedDataset` and `preprocess_data.py`.
- I have completely redesigned the preprocessing script, simplifying it to the maximum. Now the user can use both Hugging Face Datasets and JSON files similar to those used in other projects like Megatron. Basically, once we build the Dataset, we shard it among the number of workers. Each worker will be responsible for tokenizing the text present in its shard and storing it in a _numpy array_. Finally, we concatenate all the arrays produced by each worker. This way, we get rid of the trick to import nanotron without installing it, revert to the original `log_rank` function (which we had modified because it required initializing the distributed group) but, as we are now working with Hugging Face Datasets, we need to add Datasets as a project requirement. One last note: We store the tokens with `dtype = np.uint16`, which limits us to using tokenizers with vocabularies < 65536. We could change it to int32, although it is true that we would double the space.

- I have also redesigned the `MMapIndexedDataset`, the heart of the `Nanosets`. Now it is much simpler, as it only contains the `np.memmap` to read the tokens. It contains the `__getitem__` method, although the important one is the `get`, which allows us to access an `offset` position of the array and extract `length` tokens. We don't need more for CausalLM.

And where did all the sample and document index things from the `Nanoset` go? Actually, we don't need them, as the only thing that interests us is to extract `sequence length + 1` tokens from the mmap dataset. This way, we save a lot of time in the process of building the indices. Now we simply focus on how many tokens our mmap array has, based on the number of tokens and the sequence length we compute the total number of samples that we can generate from the array and divide them into train, valid, and test.

What we do keep to access the `Nanoset` indices correctly is the shuffle index and the same logic we used before to build them (Concatenating them in the train split to ensure that we have enough samples), since it is worth remembering that the 3 splits access the same `MMapIndexedDataset`, but at different positions.

I have updated the docs and here I have uploaded the logs with the same configuration that I published a few days ago. The only thing that changes is the number of samples per Nanoset split, as we now do the division at the token level and not at the document level. I have also added a new test to verify that we are able to reset the state of the dataloader when restarting training.

Toni

tests/test_build_nanoset_dataloader.py
```diff
@@ -0,0 +1,161 @@
...
+            micro_batch_size=MICRO_BATCH_SIZE,
+            dataloader_num_workers=0,
+            dataloader_drop_last=True,
+            # NOTE The dataloader serves batches of micro_batch_size despite of batch_accumulation_per_replica
+            consumed_train_samples=SKIPED_BATCHES * MICRO_BATCH_SIZE * parallel_context.dp_pg.size(),
+        )
+
+        recovered_first_batch = next(iter(recovered_dataloader))
+
+        assert compute_batch_hash(batch) == compute_batch_hash(recovered_first_batch)
```
xrsrke, line 159: We could move this to a separate test (making each test only test a specific expected behavior).

tests/helpers/data.py
```diff
@@ -0,0 +1,203 @@
...
+    """
+    Creates a sha256 hash from the elements of a OrderedDict
+    """
+    unique_description = json.dumps(identifier, indent=4)
+    # Create n_digit description hash
+    unique_description_hash = int(hashlib.sha256(unique_description.encode("utf-8")).hexdigest(), 16) % 10**n_digit
+    return unique_description_hash
+
+
+def assert_nanoset(nanoset: Nanoset, parallel_context: ParallelContext):
```
xrsrke, line 127: `assert_nanoset_sync_across_dp_ranks`

docs/nanoset.md
```diff
@@ -0,0 +1,161 @@
...
+</pre>
+
+The dataset is then processed into a mmap format for training using the [`tools/preprocess_data.py`](../tools/preprocess_data.py) script. Below we show an example for processing a corpus with the Llama2 tokenizer.
+
+<pre>
+python tools/preprocess_data.py \
+       --input data/my_corpus.json \
+       --output-prefix data/processed-datasets/my-llama2-dataset \
+       --pretrained-model-name-or-path models/Llama-2-7b \
+       --workers 128
```
xrsrke, line 22:
```suggestion
       --num-workers 128
```

xrsrke (changes requested):
You forgot to add `NanosetDatasetsArgs` to `get_dataloader_from_data_stage` [[link]](https://github.com/huggingface/nanotron/blob/e552f3ff1a6202591a6d2c63522aaded49288248/run_train.py#L123-L124)


```
File "/temp/0320/nanotron/run_train.py", line 157, in <module>
  File "/temp/0320/nanotron/run_train.py", line 157, in <module>
Traceback (most recent call last):
  File "/temp/0320/nanotron/run_train.py", line 157, in <module>
  File "/temp/0320/nanotron/run_train.py", line 157, in <module>
        dataloader = get_dataloader(trainer)dataloader = get_dataloader(trainer)

  File "/temp/0320/nanotron/run_train.py", line 137, in get_dataloader
dataloader = get_dataloader(trainer)  File "/temp/0320/nanotron/run_train.py", line 137, in get_dataloader

  File "/temp/0320/nanotron/run_train.py", line 137, in get_dataloader
        dataloader = get_dataloader(trainer)get_dataloader_from_data_stage(trainer, stage.data)

get_dataloader_from_data_stage(trainer, stage.data)  File "/temp/0320/nanotron/run_train.py", line 137, in get_dataloader

  File "/temp/0320/nanotron/run_train.py", line 124, in get_dataloader_from_data_stage
  File "/temp/0320/nanotron/run_train.py", line 124, in get_dataloader_from_data_stage
    get_dataloader_from_data_stage(trainer, stage.data)
  File "/temp/0320/nanotron/run_train.py", line 124, in get_dataloader_from_data_stage
    get_dataloader_from_data_stage(trainer, stage.data)
  File "/temp/0320/nanotron/run_train.py", line 124, in get_dataloader_from_data_stage
    raise ValueError(f"Unhandled case of `self.config.data.dataset`. Got: {data.dataset}")
raise ValueError(f"Unhandled case of `self.config.data.dataset`. Got: {data.dataset}")
ValueError: Unhandled case of `self.config.data.dataset`. Got: NanosetDatasetsArgs(data_path='/temp/0320/nanotron/data/processed-datasets/tmp', split='949,50,1', path_to_cache=None)
ValueError: Unhandled case of `self.config.data.dataset`. Got: NanosetDatasetsArgs(data_path='/temp/0320/nanotron/data/processed-datasets/tmp', split='949,50,1', path_to_cache=None)
    raise ValueError(f"Unhandled case of `self.config.data.dataset`. Got: {data.dataset}")
ValueError: Unhandled case of `self.config.data.dataset`. Got: NanosetDatasetsArgs(data_path='/temp/0320/nanotron/data/processed-datasets/tmp', split='949,50,1', path_to_cache=None)
    raise ValueError(f"Unhandled case of `self.config.data.dataset`. Got: {data.dataset}")
ValueError: Unhandled case of `self.config.data.dataset`. Got: NanosetDatasetsArgs(data_path='/temp/0320/nanotron/data/processed-datasets/tmp', split='949,50,1', path_to_cache=None)
[2024-04-02 13:31:16,617] torch.distributed.elastic.multiprocessing.api: [ERROR] failed (exitcode: 1) local_rank: 0 (pid: 3732072) of binary: /temp/0320/env/bin/python
Traceback (most recent call last):
  File "/temp/0320/env/bin/torchrun", line 8, in <module>
    sys.exit(main())
  File "/temp/0320/env/lib/python3.10/site-packages/torch/distributed/elastic/multiprocessing/errors/__init__.py", line 347, in wrapper
    return f(*args, **kwargs)
  File "/temp/0320/env/lib/python3.10/site-packages/torch/distributed/run.py", line 812, in main
    run(args)
  File "/temp/0320/env/lib/python3.10/site-packages/torch/distributed/run.py", line 803, in run
    elastic_launch(
  File "/temp/0320/env/lib/python3.10/site-packages/torch/distributed/launcher/api.py", line 135, in __call__
    return launch_agent(self._config, self._entrypoint, list(args))
  File "/temp/0320/env/lib/python3.10/site-packages/torch/distributed/launcher/api.py", line 268, in launch_agent
    raise ChildFailedError(
torch.distributed.elastic.multiprocessing.errors.ChildFailedError:
```

TJ-Solergibert:
Hello! 

Thank you for your comments, I will review them later. Regarding what you mentioned, you have to start the training with [_run_train_nanoset.py_](https://github.com/huggingface/nanotron/blob/e3190be3406cb19bf9d4e5129bebe68d59b030c4/run_train_nanoset.py) and not _run_train.py_. If you prefer, I can merge everything.

Toni

tests/helpers/data.py
TJ-Solergibert: I changed it to `assert_nanoset_sync_across_all_ranks` because we check that in all ranks we have the same Nanoset (and the same for BlendedNanoset)

tests/test_build_nanoset_dataloader.py
TJ-Solergibert: I have divided the tests: Now we have `test_build_nanoset_dataloader`, which checks that we correctly build the datasets and DataLoaders in each process, and `test_recover_nanoset_dataloader`, which checks that when restarting training, we are able to recover the state prior to the interruption. In the case of `test_recover_nanoset_dataloader`, it is necessary to `skip batches > train steps`.

TJ-Solergibert:
Hello!

I have integrated Nanosets into `run_train.py`. I have added a yaml config file to run experiments. Previously, it will be necessary to download and preprocess two datasets. I have chosen [yelp_review_full](https://huggingface.co/datasets/yelp_review_full) and [HuggingFaceH4/testing_alpaca_small](https://huggingface.co/datasets/HuggingFaceH4/testing_alpaca_small) from the Hugging Face Hub, and we use the GPT2 Tokenizer.

We download and preprocess the datasets as follows:

```sh
python3 tools/preprocess_data.py \
       --input yelp_review_full \
       --split train \
       --output-prefix datasets/yelp_review_full \
       --pretrained-model-name-or-path gpt2 \
       --num-workers 16
```

```sh
python3 tools/preprocess_data.py \
       --input HuggingFaceH4/testing_alpaca_small \
       --split train \
       --column completion \
       --output-prefix datasets/testing_alpaca_small \
       --pretrained-model-name-or-path gpt2 \
       --num-workers 16
```

We launch the job with:

```sh
torchrun --nproc-per-node 4 run_train.py --config examples/config_nanoset.yaml
```

I have tested it with a setup with 4 GPUs.

Toni.

xrsrke (approved):
all good. thanks for the fantastic contribution!

pyproject.toml
```diff
@@ -20,7 +20,9 @@ dependencies = [
     "packaging",
     "safetensors",
     "dacite",
-    "tqdm"
+    "tqdm",
+    "transformers",
```
xrsrke, line 24: ah i forgot, we should not require transformers or datasets, because it could be optional

xrsrke (changes requested):
one last change before we merge

src/nanotron/data/dataset_builder.py
```diff
@@ -0,0 +1,209 @@
...
+        config (NanosetConfig): The config object which informs dataset creation
+    """
+
+    def __init__(
+        self,
+        config: NanosetConfig,
+    ):
+        self.config = config
+
+    def build(self) -> List[Union[BlendedNanoset, Nanoset]]:
```
NouamaneTazi, line 32: is there a reason why we have a `build()` function? Maybe we can just return the nanosets in the init call?

NouamaneTazi (changes requested):
Looking really good :) Final review before merging this!

TJ-Solergibert:
Hi! 

Throughout this week, I'll check all your recommendations, thank you! 

`transformers` and `datasets` are necessary for data processing (mainly `transformers` for the tokenizer) so I've though of creating a new flavour/extension for the Nanosets and including them to the [3d_parallelism_unit_tests.yaml](https://github.com/huggingface/nanotron/blob/de2ca7a85180f10f342473bea0dac81bb7293e14/.github/workflows/3d_parallelism_unit_tests.yaml#L47) workflow. 

Toni

pyproject.toml
TJ-Solergibert: Done! Added `nanosets` package to `pyproject.toml` and we install it in `.github/workflows/3d_parallelism_unit_tests.yaml`. Also added to `docs/nanoset.md`. Should we include something in `README.md`?

docs/nanoset.md
```diff
@@ -0,0 +1,161 @@
...
+
+<pre>
+python tools/preprocess_data.py \
+       --input data/my_corpus.json \
+       --output-prefix data/processed-datasets/my-llama2-dataset \
+       --pretrained-model-name-or-path models/Llama-2-7b \
+       --num-workers 128
+</pre>
+
+In `--pretrained-model-name-or-path`, we will have to specify a tokenizer in the same way as we do when using `AutoTokenizers.from_pretrained(...)`.
```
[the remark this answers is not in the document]
TJ-Solergibert: This refers to the tokenizer we use to preprocess the data, I don't understand your comment about `examples/train_tiny_llama.sh`. As suggested, I've changed `--pretrained-model-name-or-path` to `--tokenizer-name-or-path`

docs/nanoset.md
```diff
@@ -0,0 +1,161 @@
...
+{"src": "The Internet", "text": "jumps over the lazy dog", "type": "Eng", "id": "42", "title": "Second Part"}
+</pre>
+
+The dataset is then processed into a mmap format for training using the [`tools/preprocess_data.py`](../tools/preprocess_data.py) script. Below we show an example for processing a corpus with the Llama2 tokenizer.
+
+<pre>
+python tools/preprocess_data.py \
+       --input data/my_corpus.json \
+       --output-prefix data/processed-datasets/my-llama2-dataset \
+       --pretrained-model-name-or-path models/Llama-2-7b \
```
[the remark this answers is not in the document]
TJ-Solergibert: `--pretrained-model-name-or-path` comes from the `from_pretrained` methods of `transformers`. Changed to `--tokenizer-name-or-path`

tools/preprocess_data.py
```diff
@@ -0,0 +1,121 @@
...
+        lambda x: {"input_ids": tokenizer.encode(x, add_special_tokens=add_special_tokens)},
+        input_columns=column,
+        batched=False,
+        remove_columns=[column],
+    )
+    input_ids_file = open(output_file, "wb")
+    for sample in dataset_shard:
+        np_array = np.array(sample["input_ids"], dtype=np.uint16)
+        input_ids_file.write(np_array.tobytes(order="C"))
+    input_ids_file.close()
```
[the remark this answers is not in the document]
TJ-Solergibert: I added `batched=True`, which improves performance somewhat. Regarding the writing process: Dataset preprocessing is done in two stages: the first one where each worker `W` processes `len(dataset)/W` samples and writes the intermediate result to disk, and the second one where the main process merges all these files. In our setup with GPFS, we observed that it was faster to have W workers writing to disk and then merging the results (which is a low-cost operation) than having a single process writing all the tokens to disk. Another option I tried was writing to a memmap, but required counting the total number of tokens (For big datasets becomes unfeasible).

tests/helpers/data.py
```diff
@@ -0,0 +1,203 @@
...
+    nanoset_identifiers["random_seed"] = nanoset.config.random_seed
+    nanoset_identifiers["sequence_length"] = nanoset.config.sequence_length
+    nanoset_identifiers["length"] = len(nanoset)
+    nanoset_identifiers["input_ids"] = nanoset[IDX_SAMPLE]["input_ids"].tolist()
+    nanoset_identifiers["indices"] = nanoset.indexed_indices.tolist()
+
+    unique_description_hash = compute_hash(nanoset_identifiers)
+    assert_tensor_synced_across_pg(
+        tensor=torch.tensor(unique_description_hash, device="cuda"),
+        pg=parallel_context.world_pg,
```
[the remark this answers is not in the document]
TJ-Solergibert: No, in **all** processes, we build the **same** `Nanoset`; what changes are the indices we extract. The indices we extract come from the [`DistributedSampler`](https://github.com/huggingface/nanotron/blob/ad21be06bae6feead121f308cfa02255e0942633/src/nanotron/dataloader.py#L430), which divides the total number of indices (the dataset length) among the number of DP ranks.

src/nanotron/data/dataset_builder.py
TJ-Solergibert: `NanosetBuilder` returns a list of `Nanosets` or `BlendedNanosets` according to the `NanosetConfig`. I know it's not the most aesthetic solution, but in `__init__`, we can't return anything, so we use `__init__` to set the object's attributes and `build()` to return the constructed datasets. I tried using other methods like `__new__`, and it ended up being more complex.

TJ-Solergibert:
Don't merge this PR; I'll open a new one with a truly Nano dataset 👀

PR #103, opened 2024-03-13
Add converter (HF <-> Nanotron) for Mamba

3outeille:
- Can now convert HF Mamba model to Nanotron in every parallelism (DP/PP/TP)
- Can now convert Nanotron model in every parallelism (DP/PP/TP) to HF model


3outeille:
Ready for first review. Just need to fix the PP bug because of inference_param which is not a TensorPointer

examples/mamba/convert_hf_to_nanotron.py
```diff
@@ -0,0 +1,342 @@
...
+# def sanity_check_weights(model, model_ref, tp_size):
+#     def _sort_key(name_param_pair):
+#         name, _ = name_param_pair
+#         # Split the name and take the last part as the key for sorting
+#         return name.split('.')[-1]
+    
+#     def _split_weight(data: torch.Tensor, dim: int) -> torch.Tensor:
+#         rank = dist.get_rank()
+#         world_size = dist.get_world_size()
+#         chunks = torch.chunk(data, world_size, dim=dim)
```
NouamaneTazi, line 50: can we delete commented parts?

examples/mamba/convert_hf_to_nanotron.py
```diff
@@ -0,0 +1,342 @@
...
+        def _get_tensor(path: str):
+            param = ref_module.get_parameter(path)
+            return param.grad
+    
+    param = _get_tensor(hf_name)
+    
+    if "in_proj" in hf_name:
+        # In Nanotron, we do tensor parallel column so weight need to be split in the column dimension (i.e: xz.view(...))
+        # However, the HF weights was trained such that it expected xz.chunk(...) to split the tensor in the row dimension
+        # Thus, we need to interleaved the HF weights to make it compatible with Nanotron
```
NouamaneTazi, line 119: nice comments!

examples/mamba/convert_hf_to_nanotron.py
```diff
@@ -0,0 +1,342 @@
...
+        log_rank(f"Interleaving {hf_name} to make it compatible with Nanotron", logger=logger, level=logging.INFO, rank=0)
+        param = param[_interleave_pattern(param.shape[0]), :]
+        
+    return param
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Convert HF weights from states-space repo to brrr weights")
+    parser.add_argument("--model", type=str, default="130M", help="130M | 370M | 790M | 1.4B | 2.8B")
```
NouamaneTazi, line 127: I think it's better to have the entire model name here and not just 130M
The assert below can check if it's a states space mamba model for example

examples/mamba/mamba.py
```diff
@@ -98,6 +104,7 @@ def __init__(
         self.dt_rank = math.ceil(self.d_model / 16) if dt_rank == "auto" else dt_rank
         self.use_fast_path = use_fast_path
         self.layer_idx = layer_idx
+        self.is_inference = is_inference
```
NouamaneTazi, line 107: is_inference can be defined by checking if store has been created and not null. It's better than having a param for this imo

NouamaneTazi: or we can remove the is_inference param and just check if store is defined, because we only check for inference if we want to use the store

examples/mamba/mamba.py
```diff
@@ -439,40 +445,44 @@ def forward(self, input_ids: torch.Tensor, input_mask: torch.Tensor):  # [batch_
 class MambaDecoderLayer(nn.Module):
     def __init__(
         self,
-        config: MambaModelConfig,
+        model_config: MambaModelConfig,
```
NouamaneTazi, line 448: Hmm so far we've been calling them config (same as transformers). Can we stick to that? 🙏

examples/mamba/mamba.py
```diff
@@ -439,40 +445,44 @@ def forward(self, input_ids: torch.Tensor, input_mask: torch.Tensor):  # [batch_
...
         parallel_config: Optional[ParallelismArgs],
         tp_pg: dist.ProcessGroup,
         layer_idx: int,
         device: Optional[torch.device] = None,
         dtype: Optional[torch.dtype] = None,
+        is_inference: bool = False,
+        store: Optional[Store] = None,
```
NouamaneTazi, line 455: store shouldnt be in args. It's attachable to the module

examples/mamba/mamba.py
```diff
@@ -732,21 +753,43 @@ def forward(
...
         super().__init__()
 
+        self.parallel_context = parallel_context
+        self.model_config = model_config
+        self.parallel_config = parallel_config
+        
+        store = None
+        
+        if is_inference and inference_config is not None:
+            self._attach_store(Store())
```
NouamaneTazi, line 775: the store should be attached automatically no? just like we do for kv cache

examples/mamba/convert_hf_to_nanotron.py
```diff
@@ -0,0 +1,342 @@
...
+        "complex128": torch.complex128,
+        "float16": torch.float16,
+        "bfloat16": torch.bfloat16,
+        "uint8": torch.uint8,
+        "int8": torch.int8,
+        "int16": torch.int16,
+        "int32": torch.int32,
+        "int64": torch.int64,
+        "bool": torch.bool,
+    }
```
NouamaneTazi, line 217: we don't need this. we can just check for a single dtype. You can set dtype to bfloat16 for example and use that variable for the rest of this script

examples/mamba/convert_hf_to_nanotron.py
```diff
@@ -0,0 +1,342 @@
...
+    attrs = yaml.safe_load(yaml_content)
+    model_config = MambaModelConfig(**attrs)
+    
+    assert model_config.dtype == "float32", "Convert weights only in float32"
+
+    # Initiliaze Brrr model
+    model_config.vocab_size = _vocab_size_with_padding(
+            model_config.vocab_size,
+            pg_size=parallel_context.tp_pg.size(),
+            make_vocab_size_divisible_by=5, # So that every value of TP from 1 to 8 yield a vocab_size of 50280
```
NouamaneTazi, line 229: not sure I understand this comment. Why does TP care abt vocab size?

examples/mamba/convert_hf_to_nanotron.py
```diff
@@ -0,0 +1,342 @@
...
+        num_hidden_layers = 24
+        pretrained_model_name = "state-spaces/mamba-790m"
+    elif args.model == "1.4B":
+        d_model = 2048
+        num_hidden_layers = 48
+        pretrained_model_name = "state-spaces/mamba-1.4b"
+    elif args.model == "2.8B":
+        d_model = 2560
+        num_hidden_layers = 64
+        pretrained_model_name = "state-spaces/mamba-2.8b"
```
NouamaneTazi, line 186: here as well. I think it's better to just specify the model_name. And then load its config from transformers, then load the d_model and num_hidden_layers..

examples/mamba/run_generate.py
```diff
@@ -0,0 +1,297 @@
...
+        "complex128": torch.complex128,
+        "float16": torch.float16,
+        "bfloat16": torch.bfloat16,
+        "uint8": torch.uint8,
+        "int8": torch.int8,
+        "int16": torch.int16,
+        "int32": torch.int32,
+        "int64": torch.int64,
+        "bool": torch.bool,
+    }
```
NouamaneTazi, line 137: same as above

src/nanotron/generation/decode.py
```diff
@@ -243,12 +244,19 @@ def decode_text(
                     new_decoder_states.append(state)
                     # Get the new logits
                     if generation_config.use_cache:
-                        with attach_store(model=model, store=state.store):
-                            # transpose: [sequence_length, batch_size, vocab_size] -> [batch_size, sequence_length, vocab_size]
+                        if hasattr(model, "_store"):
+                            # Some model like mamba already has store build up for generation
```
NouamaneTazi, line 248: why not just keep attach_store and use it? we dont need to build it manually no?

examples/mamba/convert_hf_to_nanotron.py
3outeille: `_vocab_size_with_padding`does this `multiple = make_vocab_size_divisible_by * pg_size` and then `after = int(ceil(orig_vocab_size / multiple) * multiple)`.  Not sure why this function is this way but I found that 5 gives 50280 for every value of TP. Maybe it is better that I hardcode the `model_config.vocab_size=50280` directly

src/nanotron/generation/decode.py
3outeille: I need to update `store` value at `MambaModel` and `Mamba` class. Making those classes inherit `AttachableStore` doesn't share `store` across the classes. Thus, I had to create it once during `MambaForTraining` init and pass it as argument to to all the other classes

This implies the `if/else` you highlighted as in `run_generate.py` we first build the model (which create a `store` in `MambaForTraining` init) before getting into `decode_text` which create a new `store` at `with attach_store()`

Do you have any suggestion to make it cleaner ?

examples/mamba/convert_hf_to_nanotron.py
NouamaneTazi: I think we should keep the value provided by vocab_size nad have vocab_size=50280 in the config? Is there any reason why we want to force it to be a multiple of 5 though?

src/nanotron/generation/decode.py
3outeille: okay found a cleaner way, see commit [26c1f3e](https://github.com/huggingface/nanotron/pull/103/commits/26c1f3e3e35ed1b5a2c37e013fd91d6bf989c2c3)

NouamaneTazi (changes requested):

examples/mamba/mamba.py
```diff
@@ -370,30 +370,34 @@ def allocate_inference_cache(self, batch_size: int, max_seqlen: int, dtype: torc
...
                 dtype=self.dt_proj.weight.dtype,
                 # dtype=torch.float32,
             )
-            inference_params.key_value_memory_dict[self.layer_idx] = (
+            
+            store["key_value_memory_dict"][self.layer_idx] = (
                 conv_state,
-                ssm_state,
+                ssm_state
             )
```
NouamaneTazi, line 398: You can remove the layer_idx dependence here, and revert your changes to store (store should be instance dependent)

examples/mamba/mamba.py
```diff
@@ -349,7 +349,7 @@ def step(
         out = self.out_proj(y)
         return out.unsqueeze(1), conv_state, ssm_state
 
-    def allocate_inference_cache(self, batch_size: int, max_seqlen: int, dtype: torch.dtype = None, **kwargs):
+    def allocate_inference_cache(self, batch_size: int, max_new_tokens: int, dtype: torch.dtype = None, **kwargs):
```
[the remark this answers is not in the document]
3outeille: Actually we don't use this function so we can remove it

src/nanotron/generation/decode.py
```diff
@@ -215,11 +216,24 @@ def decode_text(
...
+            if model.__class__.__name__ == "MambaModel":
+
+                store.update(
+                    {
+                        "max_new_tokens": max_new_tokens,
+                        "max_batch_size": 1, # We are processing only one token at a time
+                        "seqlen_offset": 0,
+                        "key_value_memory_dict": {},
+                    }
+                )
```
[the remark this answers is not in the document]
3outeille: > you can move this logic in the mamba modeling code like we do for kv caching
If we do that, the `seqlen_offset` in  `Mamba.forward()` won't get the increment update from `MambaModel.forward_with_hidden_states()`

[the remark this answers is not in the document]
3outeille: > you can move this logic in the mamba modeling code like we do for kv caching

Problem is the `seqlen_offset` in `Mamba.forward` won't get the increment update done in `MambaModel.forward_with_hidden_states()`

examples/mamba/mamba.py
```diff
@@ -61,12 +61,14 @@
 except ImportError:
     RMSNorm, layer_norm_fn, rms_norm_fn = None, None, None
 
-# import lovely_tensors as lt; lt.monkey_patch()
-
 logger = logging.get_logger(__name__)
 
+from dataclasses import dataclass, field
+from torch import Tensor
+from nanotron.parallel.sharded_parameters import SplitConfig, create_sharded_parameter_from_config
```
NouamaneTazi, line 68: move before logger. It's weirf that pre-commit doesnt fix this for you

examples/mamba/mamba.py
```diff
@@ -195,21 +194,27 @@ def __init__(
...
-            conv_state, ssm_state = self._get_states_from_cache(inference_params, batch)
-            if inference_params.seqlen_offset > 0:
+        
+        store = self.get_local_store()
+        if store is not None:
+            
+            if "key_value_memory_list" not in store:
+                store["key_value_memory_list"] = []
+            
+            if "seqlen_offset" not in self._store:
```
NouamaneTazi, line 212: Why do we sometimes use self._store?

examples/mamba/mamba.py
```diff
@@ -195,21 +194,27 @@ def __init__(
...
+                self._store["seqlen_offset"] = 0
+            
+            conv_state, ssm_state = self._get_states_from_cache(batch)
```
NouamaneTazi, line 215: batch -> batch_size

examples/mamba/mamba.py
```diff
@@ -349,51 +357,33 @@ def step(
...
-            inference_params.key_value_memory_dict[self.layer_idx] = (
+            store["key_value_memory_list"] = (
                 conv_state,
-                ssm_state,
+                ssm_state
             )
         else:
-            conv_state, ssm_state = inference_params.key_value_memory_dict[self.layer_idx]
+            conv_state, ssm_state = store["key_value_memory_list"]
             # TODO: What if batch size changes between generation, and we reuse the same states?
```
NouamaneTazi, line 387: In a single generation call we can assume batchsize stays the same (so batch_size doesnt change for this store)

examples/mamba/mamba.py
```diff
@@ -349,51 +357,33 @@ def step(
...
                 dtype=self.conv1d.weight.dtype,
             )
             ssm_state = torch.zeros(
                 batch_size,
-                self.d_model * self.expand,
+                self.d_model * self.expand // self.tp_pg.size(),
                 self.d_state,
                 device=self.dt_proj.weight.device,
                 dtype=self.dt_proj.weight.dtype,
                 # dtype=torch.float32,
```
NouamaneTazi, line 379: do we need it to be in fp32 or not?

examples/mamba/mamba.py
```diff
@@ -635,6 +622,9 @@ def forward_with_hidden_states(
         sharded_logits = self.lm_head(x=hidden_states)["logits"]
         fp32_sharded_logits = self.cast_to_fp32(x=sharded_logits)["output"]
 
+        if self._store is not None:
+            self._store["seqlen_offset"] += 1
```
NouamaneTazi, line 626: why do we update this here? Let's update store only in the classes which use it

NouamaneTazi (commented):
Nice refactor! Left some more comments

3outeille:
should be done now

examples/mamba/config.py
```diff
@@ -5,14 +5,12 @@
 
 from nanotron.config import Config, ExistingCheckpointInit, NanotronConfigs
 from nanotron.config.utils_config import cast_str_to_torch_dtype
-
-
 @dataclass
 class MambaInit:
     # mamba_ssm.models.mixer_seq_simple._init_weights
```
NouamaneTazi, line 10: clean comments?

examples/mamba/config.py
```diff
@@ -5,14 +5,12 @@
...
     initializer_range: float = 0.02
-    rescale_prenorm_residual: bool = (True,)
-    n_residuals_per_layer: int = (1,)  # Change to 2 if we have MLP
+    rescale_prenorm_residual: bool = True
+    n_residuals_per_layer: int = 1  # Change to 2 if we have MLP
```
NouamaneTazi, line 13: maybe remove for now as we dont support MLP

examples/mamba/mamba.py
```diff
@@ -170,18 +166,31 @@ def __init__(
         # Our initialization would set all Linear.bias to zero, need to mark this one as _no_reinit
         self.dt_proj.bias._no_reinit = True
```
NouamaneTazi, line 167: is this needed?

src/nanotron/generation/decode.py
```diff
@@ -166,6 +166,7 @@ def decode_text(
     max_micro_batch_size: int,
     max_new_tokens: int,
     is_bench: bool = False,
+    is_logits_transpose: bool = True,
```
NouamaneTazi, line 169: maybe rename to logits_are_batch_first or logits_are_sequence_first

examples/mamba/convert_hf_to_nanotron.py
```diff
@@ -0,0 +1,264 @@
...
+        nanotron_to_hf[f'decoder.{i}.pp_block.mixer.D'] = f'backbone.layers.{i}.mixer.D'
+        nanotron_to_hf[f'decoder.{i}.pp_block.mixer.in_proj.weight'] = f'backbone.layers.{i}.mixer.in_proj.weight'
+        nanotron_to_hf[f'decoder.{i}.pp_block.mixer.conv1d.weight'] = f'backbone.layers.{i}.mixer.conv1d.weight'
+        nanotron_to_hf[f'decoder.{i}.pp_block.mixer.conv1d.bias'] = f'backbone.layers.{i}.mixer.conv1d.bias'
+        nanotron_to_hf[f'decoder.{i}.pp_block.mixer.x_proj.weight'] = f'backbone.layers.{i}.mixer.x_proj.weight'
+        nanotron_to_hf[f'decoder.{i}.pp_block.mixer.x_proj.bias'] = f'backbone.layers.{i}.mixer.x_proj.bias'
+        nanotron_to_hf[f'decoder.{i}.pp_block.mixer.dt_proj.weight'] = f'backbone.layers.{i}.mixer.dt_proj.weight'
+        nanotron_to_hf[f'decoder.{i}.pp_block.mixer.dt_proj.bias'] = f'backbone.layers.{i}.mixer.dt_proj.bias'
+        nanotron_to_hf[f'decoder.{i}.pp_block.mixer.out_proj.weight'] = f'backbone.layers.{i}.mixer.out_proj.weight'
+        #TODO: Maybe check if bias exists?
```
NouamaneTazi, line 199: del

examples/mamba/convert_hf_to_nanotron.py
```diff
@@ -0,0 +1,264 @@
+# ruff: noqa: E402
+"""
+Converts a HF model from (https://huggingface.co/state-spaces/) to a Brrr model
+
+Command:
+    torchrun --nproc_per_node=1 convert_hf_to_nanotron.py --model 130M  --save_path nanotron-weights
```
NouamaneTazi, line 6: update docs

examples/mamba/convert_hf_to_nanotron.py
```diff
@@ -0,0 +1,264 @@
...
+    if dist.get_rank() == 0:
+        with open(save_path / "config.yaml", "w") as f:
+            config = MambaConfig(
+                general=GeneralArgs(project="test", run="mamba"),
+                parallelism=parallel_config,
+                model=ModelArgs(
+                        init_method=MambaInit(),
+                        model_config=model_config,
+                    ),
+                tokenizer=TokenizerArgs(args.model + "-hf"),
```
NouamaneTazi, line 257: test without hf / maybe use save_path

examples/mamba/convert_nanotron_to_hf.py
```diff
@@ -0,0 +1,200 @@
...
+from nanotron.models import build_model, init_on_device_and_dtype
+from nanotron.parallel import ParallelContext
+from nanotron.serialize import load_weights
+from nanotron.trainer import mark_tied_parameters
+from transformers import AutoTokenizer, MambaConfig, MambaForCausalLM
+
+logger = logging.get_logger(__name__)
+
+TOKENIZER_NAME = "state-spaces/mamba-130m-hf"
+HARCODED_PROMPT = "Hello"
```
NouamaneTazi, line 29: maybe better prompt

examples/mamba/convert_nanotron_to_hf.py
```diff
@@ -0,0 +1,200 @@
...
+)
+from nanotron.models import build_model, init_on_device_and_dtype
+from nanotron.parallel import ParallelContext
+from nanotron.serialize import load_weights
+from nanotron.trainer import mark_tied_parameters
+from transformers import AutoTokenizer, MambaConfig, MambaForCausalLM
+
+logger = logging.get_logger(__name__)
+
+TOKENIZER_NAME = "state-spaces/mamba-130m-hf"
```
NouamaneTazi, line 28: make this an arguments, and save tokenizer in the same folder

examples/mamba/run_generate.py
```diff
@@ -0,0 +1,256 @@
...
+            elif getattr(model.config, "eos_token_id", None) is not None:
+                tokenizer.pad_token_id = int(model.config.eos_token_id)
+            else:
+                tokenizer.add_special_tokens({"pad_token": "[PAD]"})
+        tokenizer.padding_side = "left"
+        tokenizer.truncation_side = "left"  # TODO @nouamane: do we want this?
+        dummy_inputs = [
+            # "Passage: Daniel went back to the garden. Mary travelled to the kitchen. Sandra journeyed to the kitchen. Sandra went to the hallway. John went to the bedroom. Mary went back to the garden. Where is Mary?\nAnswer:",
+            # "This film was probably inspired by Godzilla",
+            "Hello"
```
NouamaneTazi, line 169: bigger prompt

NouamaneTazi (commented):
LGTM! Great job 🚀

examples/mamba/mamba.py
```diff
@@ -239,7 +254,7 @@ def forward(self, hidden_states: Union[torch.Tensor, TensorPointer], inference_p
             )
         else:
             assert self.d_inner % self.tp_pg.size() == 0
-            x, z = xz.view(batch, self.d_inner // self.tp_pg.size(), 2, seqlen).chunk(2, dim=2)
+            x, z = xz.view(batch_size, self.d_inner // self.tp_pg.size(), 2, seqlen).chunk(2, dim=2)
             x = x.squeeze(2)
```
3outeille, line 258: should use flatten(start_dim=) instead ?

3outeille:
going to split this PR into multiple PR as it has a lot changes

NouamaneTazi (approved):
LGTM!!

PR #109, opened 2024-03-16
bugfix bad condition

jordane95:
This line raises error when debugging before set rank for PipelineBlock. The logic here is flawed because both True and False are not None.


xrsrke:
Thanks for the fix!

PR #111, opened 2024-03-17
Add FLOPs equations for Mamba and fix number of parameters

staghado:
This PR  does two things :

1. Fix the number of parameters calculation in `create_config_mamba.py`
2. Add `model_flops_per_s` in `get_flops_per_sec` method of the MambaModel class.

Here is a simple comparison between the formulae used in `get_flops_per_sec`(Estimated FLOPs) and the PyTorch `flop_counter` (Exact FLOPs) utility for different model sizes. The x axis is the `d_model` and the y axis is TFLOPS.

Screenshot 2024-03-17 at 20 35 41

### DIFF
--- a/examples/mamba/create_config_mamba.py
+++ b/examples/mamba/create_config_mamba.py
@@ -80,7 +80,7 @@ norm_f = model_config.d_model
 num_params = human_format(
     (
         tie_embedding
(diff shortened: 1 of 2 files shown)


3outeille:
LGTM ! thank's for your contribution !

PR #116, opened 2024-03-26
[Feature] Use `uv` instead of `pip` in CI/CD

xrsrke:


NouamaneTazi (approved):
LGTM! thanks

NouamaneTazi:
any reason why is it slower than before? can we make uv use cache as well?

image

image

PR #128, opened 2024-04-05
[Bug] Fix DoReMi's DRO Loss

xrsrke:


.gitignore
```diff
@@ -162,3 +162,4 @@ cython_debug/
 .vscode
 
 checkpoints/
+wandb/*
```
3outeille, line 165: shouldn't be push to main

src/nanotron/trainer.py
```diff
@@ -264,13 +261,35 @@ def post_training(self):
...
+        if not hasattr(self.config, "data_stages") or self.config.data_stages is None:
+            if self.current_dataloader is None:
+                if isinstance(dataloaders, tuple):
+                    dataloader = dataloaders[0]
+                else:
+                    dataloader = dataloaders
+                self.current_dataloader = sanity_check_dataloader(
+                    dataloader=dataloader, parallel_context=self.parallel_context, config=self.config
+                )
+            return
```
3outeille, line 286: return is not necessary here

src/nanotron/trainer.py
```diff
@@ -264,13 +261,35 @@ def post_training(self):
...
+        elif isinstance(dataloaders, Generator):
+            # TODO(xrsrke): this is a hacky way to handle DoReMi's dataloader
+            # remove this in the next PR
+            self.current_dataloader = dataloaders
+            return
```
3outeille, line 291: same as above

src/nanotron/trainer.py
```diff
@@ -264,13 +261,35 @@ def post_training(self):
...
+                else:
+                    dataloader = dataloaders
+                self.current_dataloader = sanity_check_dataloader(
+                    dataloader=dataloader, parallel_context=self.parallel_context, config=self.config
+                )
+            return
+        elif isinstance(dataloaders, Generator):
+            # TODO(xrsrke): this is a hacky way to handle DoReMi's dataloader
+            # remove this in the next PR
+            self.current_dataloader = dataloaders
```
3outeille, line 290: should we do sanity_check_dataloader here as well ?

.gitignore
xrsrke: this is gitignore though?

src/nanotron/trainer.py
xrsrke: nah, i stop the function execution

src/nanotron/trainer.py
xrsrke: nah, it's already a generator

PR #134, opened 2024-04-13
minor : number of gpu:s per node is not always 8

staghado:
Get number of GPU:s per node using `device_count()` instead of hardcoding it.


xrsrke (approved):
LGTM!

PR #142, opened 2024-04-19
Haojun/inference

zzhhjjj:
use the config in the run_generation.py script as we did in other scripts. Add a for_inference argument to support fixed outcome inference

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -162,3 +162,7 @@ cython_debug/
 .vscode
 
 checkpoints/
+
+# wandb
+examples/wandb
+examples/contributor-guide/wandb
--- a/run_generate.py
+++ b/run_generate.py
@@ -13,10 +13,22 @@ import os
 from pathlib import Path
 
 import torch
+
 from nanotron import distributed as dist
 from nanotron import logging
-from nanotron.config import GenerationArgs, LoggingArgs, ParallelismArgs, get_config_from_file
-from nanotron.generation.decode import GenerationInput, TokenizerConfig, decode_text, decode_tokenized
+from nanotron.config import (
+    Config,
+    GenerationArgs,
+    LoggingArgs,
+    ParallelismArgs,
+    get_config_from_file,
+)
+from nanotron.generation.decode import (
+    GenerationInput,
+    TokenizerConfig,
+    decode_text,
+    decode_tokenized,
+)
 from nanotron.logging import log_rank, set_logger_verbosity_format
 from nanotron.models import build_model
 from nanotron.parallel import ParallelContext
@@ -32,9 +44,7 @@ from nanotron.random import (
     get_synced_random_state,
     set_random_seed,
 )
-from nanotron.serialize import (
-    load_weights,
-)
+from nanotron.serialize import load_weights
 from nanotron.trainer import CONFIG_TO_MODEL_CLASS, mark_tied_parameters
 
 try:
@@ -60,17 +70,21 @@ def main():
 
     assert args.ckpt_path.exists(), f"Checkpoint path {args.ckpt_path} does not exist"
 
-    config = get_config_from_file((args.ckpt_path / "config.yaml").as_posix())
+    config = get_config_from_file((args.ckpt_path / "config.yaml").as_posix(), config_class=Config)
+    # config = get_config_from_file((args.ckpt_path / "config.yaml").as_posix())
     model_config = config.model.model_config
     tokenizer_path = config.tokenizer.tokenizer_name_or_path
 
     parallel_config = ParallelismArgs(
-        dp=args.dp or config.parallelism.dp,
-        pp=args.pp or config.parallelism.pp,
-        tp=args.tp or config.parallelism.tp,
+        # dp=args.dp or config.parallelism.dp,
+        # pp=args.pp or config.parallelism.pp,
+        # tp=args.tp or config.parallelism.tp,
+        dp=config.parallelism.dp,
+        pp=config.parallelism.pp,
+        tp=config.parallelism.tp,
         pp_engine=OneForwardOneBackwardPipelineEngine(),
-        tp_mode=TensorParallelLinearMode.ALL_REDUCE,
-        tp_linear_async_communication=True,
+        tp_mode=TensorParallelLinearMode.REDUCE_SCATTER,
+        tp_linear_async_communication=config.parallelism.tp_linear_async_communication,
     )
 
     # Initialise all process groups
--- a/run_generate.py
+++ b/run_generate.py
@@ -71,14 +71,10 @@ def main():
     assert args.ckpt_path.exists(), f"Checkpoint path {args.ckpt_path} does not exist"
 
     config = get_config_from_file((args.ckpt_path / "config.yaml").as_posix(), config_class=Config)
-    # config = get_config_from_file((args.ckpt_path / "config.yaml").as_posix())
     model_config = config.model.model_config
     tokenizer_path = config.tokenizer.tokenizer_name_or_path
 
     parallel_config = ParallelismArgs(
-        # dp=args.dp or config.parallelism.dp,
-        # pp=args.pp or config.parallelism.pp,
-        # tp=args.tp or config.parallelism.tp,
         dp=config.parallelism.dp,
         pp=config.parallelism.pp,
         tp=config.parallelism.tp,
--- a/src/nanotron/config/models_config.py
+++ b/src/nanotron/config/models_config.py
@@ -40,6 +40,7 @@ class LlamaConfig:
     tie_word_embeddings: bool = False
     use_cache: bool = True
     vocab_size: int = 32000
+    for_inference: bool = False  # if true, replace TritonRMSNorm with LayerNorm for a fixed output. use TritonRMSNorm for training as it's faster
 
     def __post_init__(self):
         # for backward compatibility
--- a/src/nanotron/models/llama.py
+++ b/src/nanotron/models/llama.py
@@ -17,6 +17,8 @@
 from typing import Dict, Optional, Union
 
 import torch
+
+# from torch.nn import LayerNorm
 from flash_attn import bert_padding
 from flash_attn.flash_attn_interface import (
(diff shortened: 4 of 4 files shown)


.gitignore
```diff
@@ -162,3 +162,7 @@ cython_debug/
 .vscode
 
 checkpoints/
+
+# wandb
```
3outeille, line 166: **/wandb/** instead

run_generate.py
```diff
@@ -13,10 +13,22 @@
 from pathlib import Path
 
 import torch
+
 from nanotron import distributed as dist
 from nanotron import logging
-from nanotron.config import GenerationArgs, LoggingArgs, ParallelismArgs, get_config_from_file
-from nanotron.generation.decode import GenerationInput, TokenizerConfig, decode_text, decode_tokenized
+from nanotron.config import (
```
3outeille, line 19: use linter as [here](https://github.com/huggingface/nanotron?tab=readme-ov-file#installation)

src/nanotron/models/llama.py
```diff
@@ -17,6 +17,8 @@
 from typing import Dict, Optional, Union
 
 import torch
+
+# from torch.nn import LayerNorm
```
3outeille, line 21: remove comment

run_generate.py
zzhhjjj: should be good for now

zzhhjjj:
need to add something

src/nanotron/models/llama.py
```diff
@@ -46,6 +53,23 @@
...
+
+    def forward(self, x):
+        output = self._norm(x.float())
+        # Llama does x.to(float16) * w whilst Gemma is (x * w).to(float16)
+        # See https://github.com/huggingface/transformers/pull/29402
+        output = output * (1.0 + self.weight.float())
+        return output.type_as(x)
+
+
 class RotaryEmbedding(nn.Module):
```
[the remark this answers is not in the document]
zzhhjjj: Didn't get it. I didn't change the rotary embedding.

.gitignore
```diff
@@ -162,4 +162,6 @@ cython_debug/
 .vscode
 
 checkpoints/
-wandb/
+
+# wandb
```
NouamaneTazi, line 166:
```suggestion
```

src/nanotron/config/config.py
```diff
@@ -401,7 +406,7 @@ def get_config_from_dict(
                 SamplerType: lambda x: SamplerType[x.upper()],
             },
             # strict_unions_match=True,
-            strict=True,
+            strict=not igonore_all_unused_keys,
```
NouamaneTazi, line 409: why do we need this? we'd like it to stay strict if possible?

src/nanotron/config/models_config.py
```diff
@@ -50,6 +50,7 @@ class LlamaConfig:
     tie_word_embeddings: bool = False
     use_cache: bool = True
     vocab_size: int = 32000
+    for_inference: bool = False  # if true, replace TritonRMSNorm with LayerNorm for a fixed output. use TritonRMSNorm for training as it's faster
```
NouamaneTazi, line 53: i don't think we should have that in the model config. just in the `run_generate.py` file is fine

src/nanotron/models/llama.py
```diff
@@ -607,15 +624,23 @@ def __init__(
         layer_idx: int,
     ):
         super().__init__()
-        self.input_layernorm = TritonRMSNorm(config.hidden_size, eps=config.rms_norm_eps)
+        # TritonRMSNorm is faster but generate randomized results. Use TritonRMSNorm for training as it's faster
+        if not config.for_inference:
+            self.input_layernorm = TritonRMSNorm(config.hidden_size, eps=config.rms_norm_eps)
+        else:
+            # replace TritonRMSNorm with RMSNorm for a fixed output.
```
NouamaneTazi, line 631:
```suggestion
            # replace TritonRMSNorm with RMSNorm for a deterministic output.
```

NouamaneTazi: replace in all ur PR

NouamaneTazi (changes requested):

src/nanotron/config/config.py
zzhhjjj: So by default it's true unless we want it to not be strict. Because, sometimes I've to either delete part of the config files by hand or change this argument.

zzhhjjj: Because in my case, I have a model trained with an older dataset config, I cannot load it with the new data_stages config and dataset config. However for evaluation, I don't care about it, I only want to load the model. This is just to add some backward compatibility, and the default value of strict is set to True so it will not trigger any bugs for anyone who is unconscious about this new key.  Btw, we already have similar arguments that can skip unused keys. https://github.com/huggingface/nanotron/blob/4799d247026a8530face56b9daec4b210732fd3b/src/nanotron/config/config.py#L385-L386

src/nanotron/config/models_config.py
zzhhjjj: I plan to add it as an optional argument in the LlamaModel and LlamaModel class __init__ function.

src/nanotron/config/config.py
zzhhjjj: An example would be: 

The older version of OptimizerArgs: 
```
class OptimizerArgs:
    ...
   adam_beta2: float
```
The latest version of OptimizerArgs:
```
class OptimizerArgs:
    optimizer_factory: Union[SGDOptimizerArgs, AdamWOptimizerArgs]
class AdamWOptimizerArgs:
    ...
    adam_beta2: float
```
For evaluation, I don't think we care about whether we load the optimizer state correctly. You are right, we'd like it to stay strict for training, and this argument should only be used for evaluation.

src/nanotron/models/llama.py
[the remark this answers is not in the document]
xrsrke: Sorry if it wasn't clear. I mean like instead of putting RotaryEmbedding for inference and training in one class, we should split it to RotaryEmbeddingForInference, and RotaryEmbeddingForTraining, then based on the config, we call the corresponding one in CausalSelfAttention.

src/nanotron/config/models_config.py
zzhhjjj: done

.gitignore
```diff
@@ -162,4 +162,5 @@ cython_debug/
 .vscode
 
 checkpoints/
-wandb/
+
+**/wandb/**
```
NouamaneTazi, line 166: i think these are the same no?

run_generate.py
```diff
@@ -68,17 +69,17 @@ def main():
...
-        dp=args.dp or config.parallelism.dp,
-        pp=args.pp or config.parallelism.pp,
-        tp=args.tp or config.parallelism.tp,
+        dp=config.parallelism.dp,
+        pp=config.parallelism.pp,
+        tp=config.parallelism.tp,
         pp_engine=OneForwardOneBackwardPipelineEngine(),
-        tp_mode=TensorParallelLinearMode.ALL_REDUCE,
-        tp_linear_async_communication=True,
+        tp_mode=TensorParallelLinearMode.REDUCE_SCATTER,
```
NouamaneTazi, line 81: keep ALL_REDUCE because sometimes REDUCE_SCATTER assumes that sequence length can be divided by TP

run_generate.py
```diff
@@ -68,17 +69,17 @@ def main():
...
     model_config = config.model.model_config
     tokenizer_path = config.tokenizer.tokenizer_name_or_path
 
     parallel_config = ParallelismArgs(
-        dp=args.dp or config.parallelism.dp,
-        pp=args.pp or config.parallelism.pp,
-        tp=args.tp or config.parallelism.tp,
+        dp=config.parallelism.dp,
+        pp=config.parallelism.pp,
+        tp=config.parallelism.tp,
```
NouamaneTazi, line 79: Revert. we want to be able to set our own dp/tp/pp

run_generate.py
```diff
@@ -68,17 +69,17 @@ def main():
...
         pp_engine=OneForwardOneBackwardPipelineEngine(),
-        tp_mode=TensorParallelLinearMode.ALL_REDUCE,
-        tp_linear_async_communication=True,
+        tp_mode=TensorParallelLinearMode.REDUCE_SCATTER,
+        tp_linear_async_communication=config.parallelism.tp_linear_async_communication,
```
NouamaneTazi, line 82: this should be fixed to False

src/nanotron/config/config.py
```diff
@@ -416,8 +419,7 @@ def get_config_from_dict(
                 RecomputeGranularity: lambda x: RecomputeGranularity[x.upper()],
                 SamplerType: lambda x: SamplerType[x.upper()],
             },
-            # strict_unions_match=True,
-            strict=True,
+            strict_unions_match=True,
```
NouamaneTazi, line 422: why? are you sure this doesn't break anything?

src/nanotron/models/llama.py
```diff
@@ -45,6 +45,28 @@
...
+            self.eps = eps
+            self.weight = nn.Parameter(torch.zeros(hidden_size))
+
+        def _norm(self, input: torch.Tensor) -> torch.Tensor:
+            return input * torch.rsqrt(input.pow(2).mean(-1, keepdim=True) + self.eps)
+
+        def forward(self, input: torch.Tensor) -> torch.Tensor:
+            output = self._norm(input.float())
+            output = output * (1.0 + self.weight.float())
+            return output.type_as(input)
```
NouamaneTazi, line 68: nice! did you check that this implementation of RMSNorm gives correct generations btw?
You can try to run lighteval evals with a llama2 model and compare to leaderboard

NouamaneTazi (changes requested):

src/nanotron/config/config.py
zzhhjjj: reverted it. I did some experiments and forgot about it.

PR #148, opened 2024-04-23
[Bug] Fix missing `.get_named_params_without_weight_decay()` in llama

xrsrke:


examples/mamba/mamba.py
```diff
@@ -813,6 +805,14 @@ def forward(
             label_mask=label_mask,
         )["loss"]
         return {"loss": loss}
+    
+    def get_named_params_without_weight_decay(self):
```
3outeille, line 809: why is it inside masked_mean?

xrsrke: ?

https://github.com/huggingface/nanotron/blob/2f9e7d8884202c7b72294aacc1d55dfe1cd43100/examples/mamba/mamba.py#L809-L815

xrsrke: it's in the NanotronModel

3outeille: sorry was reviewing through github (displayed was buggy). Anyway, we could have kept it in `MambaModel` no ?

3outeille:
lgtm

PR #152, opened 2024-04-25
Checkpoint 1.3 backwards compatibility

AleHD:
This pull request introduces three main additions:
- [x] 1.3 weights are now backwards compatible, meaning we can load checkpoints that had the `.safetensors` suffix incorrectly placed.
- [x] The logic used to get the filenames of weights shards was moved from `serialize.weights.load_weights` to `serialize.utils.get_path` thanks to the addition of `return_all_matches` keyword.
- [x] Addition of a test that verifies the capability to load weights when changing parallelism topology.

### DIFF
--- a/src/nanotron/constants.py
+++ b/src/nanotron/constants.py
@@ -2,7 +2,7 @@ import platform
 
 from packaging.version import Version, parse
 
-CHECKPOINT_VERSION = Version("1.3")
+CHECKPOINT_VERSION = Version("1.4")
 
 PY_VERSION = parse(platform.python_version())
 
--- a/src/nanotron/serialize/utils.py
+++ b/src/nanotron/serialize/utils.py
@@ -37,7 +37,6 @@ def get_path(
     suffix = tensor_name.split(".")
     suffix_path, suffix_name = suffix[:-1], suffix[-1]
 
-    suffix_name = f"{type.value}_{suffix_name}.safetensors"
 
     if exp_tp_pp_rank_and_size:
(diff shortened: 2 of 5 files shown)


zzhhjjj:
Hello,

Thank you for your PR.

As I understand it, you are verifying the checkpoint version. If it is 1.3, a different pattern will be used to retrieve the shards; otherwise, the existing method will continue. I suspect that this approach may not resolve all the issues, given that the checkpoint version 1.3 was only added to the main branch five days ago (see [commit](https://github.com/huggingface/nanotron/commit/bdb9d3d30e8361ac2761b772bfae83e7aad50a4e#diff-c70f17076eddb32ed9ef704cf37748f5725ac7ce182cafc9e7301f605059b131)), whereas the bug was introduced a month and a half ago in this [commit](https://github.com/huggingface/nanotron/commit/ad7e4421294897583122cd00d9bfa236c6d2038b#diff-d5dd91d2d5545e9d407f5da03fd9ce882b5ebe37f1573297b8c6d5a815fe905f). Additionally, checking for version 1.2 seems problematic as it might disrupt the loading logic for all checkpoints before this commit.

Timeline:
Checkpoint version=1.2 -> bug introduced -> Checkpoint version=1.3

We have already merged a quick fix that utilized a script to modify the file name: PR #151. Could you build upon this solution? Perhaps making it automatic when this pattern is detected, instead of requiring users to run the script manually.

I really appreciate your contribution.

AleHD:
Thanks for the feedback. I implemented the suggested changes and tested it locally. Seems to be working ok. Let me know if you have any other comments.

PR #155, opened 2024-04-27
Adding Nanoset dataset

TJ-Solergibert:
Hi!

After all your comments in the previous [PR](https://github.com/huggingface/nanotron/pull/102), I decided to give one last (big) overhaul to the Nanosets, and I think now we can indeed call them Nanosets. In short, I completely got rid of the `NanosetBuilder`, `NanosetConfig`, `BlendedNanoset`, and `MMapIndexedDataset`. Also, I got rid of the different Nanosets for each train, valid, and test split (I've seen code snippets where you mention that you intend to support this, so when the time comes, I'll include it). Now there are ONLY the `Nanosets`, for which we ONLY have to specify the paths to the preprocessed datasets and, if we want, a weight for each one.

Basically, for each `dataset_path` specified in the configuration, we will read the `total number of tokens` and divide it by the `sequence length` to know the `number of samples` we can construct. If we have more than one dataset and do not specify the weight for each of them, we will consume them entirely for each epoch. Otherwise, we will build the `Nanoset` respecting these weights, consuming samples from the datasets again if necessary (i.e., if we need 1000 training samples and the weight of dataset 1 is 0.9, we will need 900 samples from this dataset. If dataset 1 only has 100 samples, we will consume it 9 times per epoch).

To determine from which dataset within the `Nanoset` and which sample to extract, we will build two indices: `dataset_index`, which will select the dataset from which to extract the sample, and `dataset_sample_index`, which will select the sample from the dataset. These two indices will be constructed using a helper function ([`build_nanoset_index_helper`](https://github.com/huggingface/nanotron/blob/e32b0799ea50dc365cf6d10eafe67ffd436e5c04/src/nanotron/data/nanoset.py#L152)) that we will compile with Numba. This function is similar to the ones included in Megatron, which were compiled in C++, but in Python, for very large datasets, it took an enormous amount of time to run. Now with Numba, we solve this problem, keeping the entire project in Python. With these two indices, we will access the specific dataset and extract the sample with `sequence length + 1` tokens.

I have updated the documentation by adding a detailed explanation of how they work, although I have left many comments in [`nanoset.py`](https://github.com/huggingface/nanotron/blob/e32b0799ea50dc365cf6d10eafe67ffd436e5c04/src/nanotron/data/nanoset.py#L13). I have also updated the tests, checking the following:
1. That we build the same `Nanoset` in ALL processes.
2. That `dataset_index` has been constructed according to the `dataset_weights`.
3. That `dataset_sample_index` does not attempt to extract samples > `len(dataset)`. 

To install the necessary dependencies, I have created a new flavor of Nanotron, so it will be necessary to install it with `pip install -e '.[nanosets]'`. You can test them as follows with the [config](https://github.com/huggingface/nanotron/blob/e32b0799ea50dc365cf6d10eafe67ffd436e5c04/examples/config_nanoset.yaml) I have added:
```sh
python3 tools/preprocess_data.py \
       --input yelp_review_full \
       --split train \
       --output-prefix datasets/yelp_review_full \
       --tokenizer-name-or-path gpt2 \
       --num-workers 16
```

```sh
python3 tools/preprocess_data.py \
       --input HuggingFaceH4/testing_alpaca_small \
       --split train \
       --column completion \
       --output-prefix datasets/testing_alpaca_small \
       --tokenizer-name-or-path gpt2 \
       --num-workers 16
```

We launch the job with:

```sh
torchrun --nproc-per-node 4 run_train.py --config examples/config_nanoset.yaml
```

I have tested it with a setup with 4 GPUs.

As always, I expect your comments!

Toni.

### DIFF
--- a/docs/nanoset.md
+++ b/docs/nanoset.md
@@ -1,0 +1,174 @@
+# Nanosets
+
+## Install
+To use `Nanosets`, it's necessary to install Nanotron with the `nanosets` flavor.
+```
+pip install -e '.[nanosets]'
+```
+
+## Data pre-processing
+
+Nanotron incorporates [`Nanosets`](../src/nanotron/data/nanoset.py), a kind of datasets based on numpy memory-mapped arrays. Permite utilizar tanto un unico dataset como varios, incluso especificando los weights de cada dataset.
+
+
+To use these datasets, first, we need to preprocess the data. The input format can either be a column of a Hugging Face Dataset or a .json file containing a text sample per line. For example:
+
+<pre>
+{"src": "www.nvidia.com", "text": "The quick brown fox", "type": "Eng", "id": "0", "title": "First Part"}
+{"src": "The Internet", "text": "jumps over the lazy dog", "type": "Eng", "id": "42", "title": "Second Part"}
+</pre>
+
+The dataset is then processed into a mmap format for training using the [`tools/preprocess_data.py`](../tools/preprocess_data.py) script. Below we show an example for processing a corpus with the Llama2 tokenizer.
+
+<pre>
+python tools/preprocess_data.py \
+       --input data/my_corpus.json \
+       --output-prefix data/processed-datasets/my-llama2-dataset \
+       --tokenizer-name-or-path meta-llama/Llama-2-7b-hf \
+       --num-workers 128
+</pre>
+
+In `--tokenizer-name-or-path`, we will have to specify a tokenizer in the same way as we do when using `AutoTokenizers.from_pretrained(...)`.
+
+The output will be one file named, in this case, `my-llama2-dataset_input_ids.npy`. We will then have to specify this file in the `data_path` field in the config file.
+
+## Working with Nanosets
+
+To work with Nanosets, we just need to configure 1 argument:
+1. `data_path`: This argument specifies the file/files that will compose the `Nanoset`. There are 3 ways to specify it:
+   1. If we specify a single path, we will create a `Nanoset` from a single dataset file.
+    ```yaml
+    data_stages:
+      - name: General purpose training (Single dataset)
+        start_training_step: 1
+        data:
+          dataset:
+            dataset_path: nanosets/SlimPajama-6B_input_ids.npy
+          num_loading_workers: 0
+          seed: 1234
+    ```
+   2. If we specify a list of paths, we will create a `Nanoset` from all the dataset files. In every epoch we will consume each and every sample from each dataset randomly.
+    ```yaml
+    data_stages:
+      - name: Second purpose training (> 1 dataset)
+        start_training_step: 15
+        data:
+          dataset:
+            dataset_path:
+            - nanoset/SlimPajama-6B_input_ids.npy
+            - nanoset/europarl_input_ids.npy
+          num_loading_workers: 0
+          seed: 1234
+    ```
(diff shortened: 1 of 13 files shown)


TJ-Solergibert:
Refractored `preprocess_data.py` script to use `torchrun` instead of multiprocessing. Now you can run:

```sh
torchrun --nproc-per-node 16 tools/preprocess_data.py \
       --input yelp_review_full \
       --split train \
       --output-prefix datasets/yelp_review_full \
       --tokenizer-name-or-path gpt2
```

```sh
torchrun --nproc-per-node 16 tools/preprocess_data.py \
       --input HuggingFaceH4/testing_alpaca_small \
       --split train \
       --column completion \
       --output-prefix datasets/testing_alpaca_small \
       --tokenizer-name-or-path gpt2
```

xrsrke:
Hello. Thanks for the PR. Looking into the PR now

docs/nanoset.md
```diff
@@ -0,0 +1,131 @@
...
+              datasets/testing_alpaca_small_input_ids.npy: 0.2
+          num_loading_workers: 0
+          seed: 1234
+    ```
+> [!IMPORTANT]
+> Remember to set the `tokenizer.tokenizer_name_or_path` in the config file to the tokenizer used to preprocess the documents and set the `model.model_config.vocab_size` accordingly.
+
+Finally, to use the `Nanosets`, launch the training with [`run_train.py`](../run_train.py).
+```shell
+torchrun --nproc-per-node 8 run_train.py --config configs/nanoset_llama2.yaml
```
xrsrke, line 81: this should be `configs/config_nanoset.yaml` ?

docs/nanoset.md
```diff
@@ -0,0 +1,131 @@
...
+
+The preprocessing is done using the [`tools/preprocess_data.py`](../tools/preprocess_data.py) script. Below we show an example for processing a corpus with the Llama2 tokenizer.
+
+<pre>
+torchrun --nproc-per-node 16 tools/preprocess_data.py \
+       --input HuggingFaceH4/testing_alpaca_small \
+       --split train \
+       --column completion \
+       --output-prefix datasets/testing_alpaca_small \
+       --tokenizer-name-or-path meta-llama/Llama-2-7b-hf
```
xrsrke, line 28: since the `configs/config_nanoset.yaml` file uses the GPT-2 tokenizer, we should also use the GPT-2 tokenizer in this example for consistency

docs/nanoset.md
```diff
@@ -0,0 +1,131 @@
...
+> [!IMPORTANT]
+> Remember to set the `tokenizer.tokenizer_name_or_path` in the config file to the tokenizer used to preprocess the documents and set the `model.model_config.vocab_size` accordingly.
+
+Finally, to use the `Nanosets`, launch the training with [`run_train.py`](../run_train.py).
+```shell
+torchrun --nproc-per-node 8 run_train.py --config configs/nanoset_llama2.yaml
+```
+
+## Under the hood
+`Nanosets` are responsible of building samples of `sequence length + 1` tokens from the preprocessed dataset files. The `dataset lengths` of each dataset will be determined by the `number of total tokens / sequence length`, discarding the last sample since its length < `sequence length`.
```
xrsrke, line 85: > discarding the last sample since its length < `sequence length`.

discarding the last sample *if* its length < `sequence length`.

xrsrke (approved):
LGTM! Ready to merge

PR #157, opened 2024-04-30
llama tests

zzhhjjj:


tests/test_llama.py
```diff
@@ -0,0 +1,176 @@
...
+                assert loss < EXPECTED_LOSS_END
+
+    process.wait()  # Wait for the process to finish
+    assert process.returncode == 0
+
+
+# also run the tiny llama example. Only want to assert it can be ran.
+def test_tiny_llama():
+    # create config file
+    cmd = f"python {TINY_LLLAMA_CREATE_CONFIG_FILE}"
```
xrsrke, line 95: Write a separate test that only creates a config yaml from the python file, then another test that runs tiny llama from a config YAML

tests/test_llama.py
```diff
@@ -0,0 +1,176 @@
...
+            print(line.decode("utf-8"), end="")
+
+    process.wait()  # Wait for the process to finish
+    assert process.returncode == 0
+
+
+if __name__ == "__main__":
+    # create config file
+    cmd = f"python {CREATE_CONFIG_FILE}"
+    process = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
```
xrsrke, line 128: Can we run it through a function like test_train_tiny_llama, so that vscode pytest can run all the tests including this one (pytest ./tests)? So we don't need to run it separately through the command line

tests/test_llama.py
```diff
@@ -0,0 +1,176 @@
...
+        assert process.returncode == 0
+    except AssertionError:
+        print("Command failed with exit code:", process.returncode)
+        exit()
+    else:
+        print("Config created successfully.")
+
+    # run training
+    # set DISABLE_FLASH_ATTENTION=1 to replace flash attention implementations
+    cmd = f'DISABLE_FLASH_ATTENTION=1 FI_PROVIDER="efa" CUDA_DEVICE_MAX_CONNECTIONS=1 torchrun --nproc_per_node={NUM_GPUS} --rdzv_endpoint=localhost:29800  {TRAIN_SCRIPT} --config-file {CONFIG_FILE}'
```
xrsrke, line 147: Use pytest parametrize with with the parallel sizes:
- tp=1, dp=1, pp=1
- tp=2, dp=2, pp=2
- tp=1, dp=2, pp=1

src/nanotron/nn/layer_norm.py
```diff
@@ -51,3 +51,20 @@ def forward(
...
+class RMSNorm(nn.Module):
+    def __init__(self, hidden_size, eps=1e-6):
+        """
+        LlamaRMSNorm is equivalent to T5LayerNorm
+        """
+        super().__init__()
+        self.weight = nn.Parameter(torch.ones(hidden_size))
+        self.variance_epsilon = eps
+
+    def forward(self, input):
```
xrsrke, line 65: add type hints

src/nanotron/nn/layer_norm.py
```diff
@@ -51,3 +51,20 @@ def forward(
             is_rms_norm=True,
             return_dropout_mask=return_dropout_mask,
         )
+
+
+class RMSNorm(nn.Module):
+    def __init__(self, hidden_size, eps=1e-6):
```
xrsrke, line 57: add type hints

src/nanotron/scaling/parametrization.py
```diff
@@ -37,6 +37,7 @@ def __init__(self, config: ModelArgs):
             TensorParallelColumnLinear: self._parametrize_column_linear,
             TensorParallelRowLinear: self._parametrize_row_linear,
             TritonRMSNorm: self._parametrize_layer_norm,
+            RMSNorm: self._parametrize_layer_norm,
```
xrsrke, line 40: add the same for SpectralMupParametrizator

.github/workflows/3d_parallelism_unit_tests.yaml
```diff
@@ -59,6 +59,7 @@ jobs:
         --durations=0 \
         --ignore tests/kernels \
         --ignore tests/fp8 \
+        --ignore tests/test_llama.py \
```
xrsrke, line 62: Can we add a flag like this https://github.com/huggingface/nanotron/blob/6a4a537054ffdd68d1d352c111cd1d93762ede95/tests/kernels/test_layer_norm.py#L8 after converting that test to a functional test, instead of running it through the command line?

src/nanotron/models/llama.py
```diff
@@ -322,8 +335,22 @@ def __init__(
...
+            self.flash_rotary_embedding = FlashRotaryEmbedding(dim=self.d_qk, interleaved=True)
+            # NOTE: Only supported for training (TODO(fmom): position_ids not supported yet)
+            # forward( self, qkv: torch.Tensor, kv)
+            self.attention = CoreAttention(
+                config,
+                parallel_config=parallel_config,
+                layer_idx=layer_idx,
+            )
+        else:
+            self.flash_rotary_embedding = RotaryEmbedding(
```
xrsrke, line 350: why don't we use CoreAttention if DISABLE_FLASH_ATTENTION is False?

src/nanotron/nn/layer_norm.py
```diff
@@ -51,3 +51,20 @@ def forward(
             is_rms_norm=True,
             return_dropout_mask=return_dropout_mask,
         )
+
+
+class RMSNorm(nn.Module):
```
xrsrke, line 56: Add a correctness unit test for RMSNorm

xrsrke (changes requested):
Good job! Left some changes

tests/test_llama.py
zzhhjjj: If my understanding is correct, we can already do it without adding a function. pytest just automatically run all the functions starting with "test".

src/nanotron/models/llama.py
zzhhjjj: Because it relies on the flash attention library. I want to disable all the function implementation from flash attention library.

tests/test_llama.py
zzhhjjj: I think people normally will just do ./train_tiny_llama.sh. So I want to keep it the same as what the script does.

.github/workflows/3d_parallelism_unit_tests.yaml
zzhhjjj: Do you mean to add  @pytest.mark.fa2 to my test? But I don't want my test to be run with the fa2 workflow.

xrsrke: ah sorry, don't add `pytest.mark.fa2` beacuse that cicd run ignores fa2 flag

tests/test_llama.py
xrsrke: > I think people normally will just do ./train_tiny_llama.sh. So I want to keep it the same as what the script does.

For unit tests, if we split it into two separate tests, it would be easier to spot which one fails. For example, if creating a YAML config from a Python file fails, but the LLaMA training does not

zzhhjjj: OK! I'll split them

tests/test_llama.py
zzhhjjj: For now, the training takes 3 hours with T4. If we add more tests, I'm afraid it'll take too much time. Nouamane said that we only care about tp for now. @NouamaneTazi 
Otherwise, I can also reduce the steps to add more parameters.

src/nanotron/scaling/parametrization.py
zzhhjjj: Nice!

NouamaneTazi:
We can disable flash attention automatically for old hardware:
```python
import torch
def supports_flash_attention(device_id):
    """Check if a GPU supports FlashAttention."""
    major, minor = torch.cuda.get_device_capability(device_id)
    
    # Check if the GPU architecture is Ampere (SM 8.x) or newer (SM 9.0)
    is_sm8x = major == 8 and minor >= 0
    is_sm90 = major == 9 and minor == 0

    return is_sm8x or is_sm90

```

Issue #46, opened 2024-01-26
[Question] Correctness of backward pass of RowLinear


ufotalent:
Hi nanotron authors. Thanks for the this work that accelerates LLM training of community. I'm recently developing a feature based on nanotron which needs some tweak on the tensor parallel / sequence parallel part. However I found something seems to be problematic:

In this code piece https://github.com/huggingface/nanotron/blob/8c1a49588d0745a6404644a86547c2dd6a63640e/src/nanotron/parallel/tensor_parallel/functional.py#L417. The backward of row linear is doing:

1. Start an async all-gather to collect unsharded grad_outputs
2. At the same time, calculate the partial grad_input using grad_output * local weights
3. After 1 finishes allgather the partial grad_input and use it as a result.

However from my understanding, this causes the grad_input to be identical cross TP shards. The correct way seems to be like this:
1. Start an async all-gather to collect unsharded grad_outputs
2. At the same time, calculate the partial grad_input using grad_output * local weights
3. After 1 finishes, calculate the remaining grad_input using grad_output fetched from other ranks by: grad_output_remaining * local_weights
4. concatenate remaining_grad_input and grad_input to form a total_grad_input 

Some basic ideas behind this is:
Row linear in TP is like:

Forward:
Rank 0: X1(B, H/2) * W1(H/2, O) = Y1(B, O) -> reduce scatter -> R1(B/2, O)
Rank 1: X2(B, H/2) * W2(H/2, O) = Y2(B, O) -> reduce scatter -> R2(B/2, O)
The sharding is like:
```
X= [X1, X2]  W= [W1]    Y= [Y1+Y2]       R= [R1]                  
                [W2]                        [R2]
```
So in backwards, theroatically
dX1 = dY1 * W1_t = concat( [dR1 * W1_t], [dR2 * W1_t])
dX2 = dY2 * W2_t = concat( [dR1 * W2_t], [dR2 * W2_t])

Current implementation seems problematic. Could you help to verify or point out my mistake? Thanks

C-TC:
Same opinion here. Current implementation of _RowLinearAsyncCommunication backward pass will produce exactly the same grad of input among all TP shards which is definitely not the expected behavior.
https://github.com/huggingface/nanotron/blob/c81551700fe77fc0e9ba75cbba4bd3d8b490557c/src/nanotron/parallel/tensor_parallel/functional.py#L439

Also, somehow the test for tensor parallel did not cover grad of input for this case:
[...]

PR #172, opened 2024-05-16
Fix _RowLinearAsyncCommunication

C-TC:
As discussed in #46, the implementation of row parallel linear backward pass seems wrong, _when TP mode is reduce-scatter and async communication is enabled_. To be more specific, the gradient of input tensor is produced by 
1. compute local slice of input gradient
2. allgather slice of input gradient

This will produce the same input gradient for all TP shards, which is not the correct behavior of Row Parallel Linear layer.
Also, the check of input gradient for row parallel is also missing in test cases. After adding the test, the result for `test_tensor_parallel.py::test_row_linear[True-TensorParallelLinearMode.REDUCE_SCATTER-4-1-1]` is as follows:
Snipaste_2024-05-16_18-37-08
Only 1/4 of the input gradient values are correct, because that's the locally computed part.

This PR does the following:
1. Fixed the backward pass of `_RowLinearAsyncCommunication`, in a similar way to forward pass of `_ColumnLinearAsyncCommunication` that overlaps communication with part of computation.
2. Add the missing test that checks the correctness of input gradient in `_RowLinearAsyncCommunication`

This bug is related to convergence and can be triggered when TP mode is reduce-scatter, and async communication is enabled (which is a common setup for users). 

Please fix this, thanks!

### DIFF
--- a/.github/workflows/fa2_unit_tests.yaml
+++ b/.github/workflows/fa2_unit_tests.yaml
@@ -39,7 +39,7 @@ jobs:
         python -c "import torch; print('torch:', torch.__version__, torch)"
         python -c "import torch; print('CUDA available:', torch.cuda.is_available())"
 
-    - name: Instal nanotron
+    - name: Install nanotron
       run: |
         python -m pip install --upgrade pip
         pip install packaging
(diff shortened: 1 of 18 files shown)


3outeille:
Thank you for your PR, there was indeed a bug in the way we compute the backward pass of RowLinear async (for every rank, each local shard of `grad_output` was multiplied by their local shard weight, which is not correct)

PR #173, opened 2024-05-18
add rope_theta config var for llama

jquesnelle:
This allows changing RoPE's theta hyperparameter for Llama models. For example, Llama 3 uses `theta = 500000` instead of the default `10000`.


xrsrke:
Thanks for the PR. Merged!

Issue #114, opened 2024-03-21
FEAT: Support 1.58-bit LLMs


younesbelkada:
Hi there! 

![image](https://github.com/huggingface/nanotron/assets/49240599/38bc4c4d-f0ec-40f1-bd57-2679c7fe03f4)

Microsoft have just released the full handbook for reproduing the 1-bit LLM paper: https://github.com/microsoft/unilm/blob/master/bitnet/The-Era-of-1-bit-LLMs__Training_Tips_Code_FAQ.pdf 

Would be exciting to see if we can have an official implementation of that paper in nanotron, and support 1-bit LLM inference directly in transformers for the models that have been trained with that method

cc @NouamaneTazi @xrsrke @3outeille @thomwolf 

cc original author:
@shumingma

xrsrke:
@younesbelkada, hey, thanks for the suggestion. I've talked with @NouamaneTazi; we agree that we will add support for 1bit later on for consumer hardware because FP8 is the coolest. You get a speedup in training (FP8 matmul, this is very important), memory reduction, and it's tested at scale (180B).... So, currently, we focus on FP8 :)

PR #180, opened 2024-05-23
Fixes : https://github.com/huggingface/nanotron/issues/114

MekkCyber:
Implementation of 1.58bit LLM with Llama following the paper & handbook released by Microsoft :

https://github.com/microsoft/unilm/blob/master/bitnet/The-Era-of-1-bit-LLMs__Training_Tips_Code_FAQ.pdf

Here are The training results on 25B tokens : 

loss_curve

cc @NouamaneTazi @xrsrke @thomwolf

### DIFF
--- a/src/nanotron/parallel/tensor_parallel/nn.py
+++ b/src/nanotron/parallel/tensor_parallel/nn.py
@@ -16,6 +16,7 @@ from typing import Optional, Tuple
 
 import torch
 from torch import nn
+import torch.nn.functional as F
 
 from nanotron import distributed as dist
 from nanotron.distributed import get_global_rank
@@ -311,6 +312,17 @@ def activation_quant(x):
     y = (x * scale).round().clamp_(-128, 127) / scale
     return y
 
+def activation_quant_inference(x):
+    """Per−token quantization to 8 bits. No grouping is needed for quantization.
+    Args:
+    x: an activation tensor with shape [n, d]
+    Returns:
+    y: a quantized activation tensor with shape [n, d]
+    """
+    scale = 127.0 / x.abs().max(dim=-1, keepdim=True).values.clamp_(min=1e-5)
+    y = (x * scale).round().clamp_(-128, 127)
+    return y, scale
+
 def weight_quant(w):
     """Per−tensor quantization to 1.58 bits.
     Args:
@@ -328,6 +340,14 @@ def normalize_last_two_dimensions(tensor, eps=1e-6):
     normalized_tensor = (tensor - mean) / (std + eps)
     
     return normalized_tensor
+
+def normalize(x, dim, eps=1e-5) : 
+    scale = dim ** (-0.5)
+    #return F.normalize(x, dim=-1) * scale
+    norm_x = x.norm(2, dim=-1, keepdim=True)
+    rms_x = norm_x * scale
+    x_normed = x / (rms_x + eps)
+    return x_normed
 
 class TensorParallelColumnLinearBitNet(nn.Linear):
     def __init__(
@@ -367,6 +387,9 @@ class TensorParallelColumnLinearBitNet(nn.Linear):
             ), f"Sum of contiguous chunks ({sum(contiguous_chunks)}) must equal to out_features ({out_features})"
         split_config = SplitConfig(split_dim=0, contiguous_chunks=contiguous_chunks)
 
+        
+        self.register_parameter('weight_scale', None)
+        # self.is_model_inference = False
         mark_all_parameters_in_module_as_sharded(
             self,
             pg=self.pg,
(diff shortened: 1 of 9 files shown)


MekkCyber:
I think it has to do with the batch size. During our latest experiment, we trained the 1.58 model on 100B tokens, and we managed to get a 2.8 loss after 25B tokens with a batch size of 1024

src/nanotron/parallel/tensor_parallel/nn.py
```diff
@@ -297,3 +298,227 @@ def forward(self, input_ids: torch.Tensor) -> torch.Tensor:
...
+            pg=self.pg,
+            split_config=split_config,
+        )
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        if not self.training : 
+            w = self.weight  # a weight tensor with shape [d, k]
+            w = w.to(torch.bfloat16)
+            w_scale = getattr(self, "weight_scale").data
+            x_norm = normalize(x, self.in_features)
```
gau-nernst, line 404: Shouldn't RMSNorm here have learnable weights and biases?

src/nanotron/parallel/tensor_parallel/nn.py
```diff
@@ -297,3 +298,227 @@ def forward(self, input_ids: torch.Tensor) -> torch.Tensor:
...
+                input=x_quant,
+                weight=w,
+                bias=self.bias,
+                group=self.pg,
+                tp_mode=self.mode,
+                async_communication=self.async_communication,
+            ) / w_scale / x_scale
+        else : 
+            w = self.weight
+            x_norm = normalize(x, self.in_features)
```
gau-nernst, line 416: Shouldn't RMSNorm here have learnable weights and biases?

hjc3613:
Hi, I fetched this pr, and fintune the llama 70B using tp=8 or pp=8, before training, I have convert the llama 70B into nanotron format using this method: https://github.com/huggingface/nanotron/pull/174, but got this error:
image
so, how to finetune the 70B model into 1.58bit from scratch? could you give me some suggestion please? thank you!

src/nanotron/parallel/tensor_parallel/nn.py
gau-nernst: @MekkCyber May I know why this is marked as resolved? From my understanding of the training tips handbook, the new RMSNorm should have learnable weights as the usual RMSNorm layers. Is there a reason you left it out here? (as well as in the PR merged in huggingface/transformers).

MekkCyber:
Hey @hjc3613 thanks for the report ! you don't have to be consitent with the convert config, it should work I will investigate that! Can you tell me from your side what is the content of models/qwen2.5-72b-instruct-nanotron/model/model/decoder/0/pp_block/MLPBitNet

hjc3613:
> Hey @hjc3613 thanks for the report ! you don't have to be consitent with the convert config, it should work I will investigate that! Can you tell me from your side what is the content of models/qwen2.5-72b-instruct-nanotron/model/model/decoder/0/pp_block/MLPBitNet

thank you very much! 
My requirement is to successfully fine-tune a 70B model to a 1.58bit model, using either Qwen2.5 70B or LLama 70B - both have almost identical model architectures. According to the training code, I need to provide a dataset and a base model in nanotron format, which will then be fine-tuned to 1.58bit for inference. However, I only have a HF version of the base model, so following PR174, I'm first converting it to nanotron format. Before conversion, I need to set dp=1, pp=1, tp=1. After conversion, I'm using the code from PR180 for training. This is what I'm currently doing - if there are any unreasonable or unnecessary steps, please advise. Thank you sincerely. To my understanding, a 70B model must have pp>1 or tp>1, as a single GPU card cannot hold the complete model.

Andrefty:
Hi @MekkCyber !
Similar to what @hjc3613 wanted to try, I am trying to fine-tune the [Llama3-8B-1.58-100B-tokens](https://huggingface.co/HF1BitLLM/Llama3-8B-1.58-100B-tokens) model on a downstream task. (I have been detailing my progress in [this](https://huggingface.co/HF1BitLLM/Llama3-8B-1.58-100B-tokens/discussions/7) Hugging Face discussion thread)

So far I think I managed to correctly unpack the weights of the model using the functions provided by you, but the [`convert_hf_to_nanotron` script from this PR](https://github.com/huggingface/nanotron/pull/174) seems to not copy the `weight_scale` layers into the nanotron-format model (even after messing around with the `quantization_config`), resulting in a model that is outputting garbage when using the `run_generate` script with it (as a sanity check).

I am unsure if this is expected and if I should continue with fine-tuning, having the model in this state. I would appreciate any help or guidance you could provide on this matter.

src/nanotron/parallel/tensor_parallel/nn.py
BradZhone: > @MekkCyber May I know why this is marked as resolved? From my understanding of the training tips handbook, the new RMSNorm should have learnable weights as the usual RMSNorm layers. Is there a reason you left it out here? (as well as in the PR merged in huggingface/transformers).

Only normalize function without learnable weights can already work well, maybe you can refer to this https://arxiv.org/abs/2505.08823

lonleyodd:
@MekkCyber 
image
Hi! I noticed in your normalization function, you're computing mean and std along dimensions (-2, -1). I'm wondering if this is intentional? Typically, LayerNorm only normalizes along the last dimension (-1). Could you clarify the reasoning behind normalizing across both hidden/batch_size and feature dimensions? Thanks!

MekkCyber:
the function is not used iirc, i used it for experiments and it didn't give better results

PR #187, opened 2024-05-28
Adding support for training chat models

TJ-Solergibert:
> [!CAUTION]
> 🚨 This is a draft, still in development, and further testing needs to be done. Feel free to leave any comments!

This PR includes everything necessary to train chat models with:
1. Sample packing
2. No cross-attention contamination between packed samples
3. Training on completions only (the answers generated by the model)

This image from @sz128 is very useful to understand 1. & 2.:
image

I am developing this feature with `axolotl`'s implementation as a reference. The current status is as follows:
### IterableDatasets
This time, I have opted for and [IterableDataset](https://github.com/TJ-Solergibert/nanotron/blob/36b1c1814bf71a5fe31f95472001ee0c2ae05ad6/src/nanotron/data/chat_dataset.py#L17) instead of a map style one. The obvious benefits are that we tokenize on the fly, which allows us to easily experiment with different models/tokenizers/chat templates and saves disk space by not storing the tokens. However, the drawbacks are:
- The division between the different DP groups is done with the `split_dataset_by_node` function, which will divide the dataset's `n_shards` among the number of DP groups. If not evenly divisible, each DP group keeps 1 sample of the dataset, skipping the other examples. This is obviously not very optimal. Also, remember that since we tokenize the data on the fly, even if we divide by `n_shards`, each DP group will produce a different amount of tokens. Thus, after X steps, one DP group will have consumed its dataset 1.5 times, while another with longer samples will have done so 0.8 times.
- Since each sample of the dataset produces a different number of tokens, it is impossible to predict how many samples from the dataset we will need to construct each batch of the dataloader. This, along with building the batches with a buffer that processes samples from the dataset until it has enough tokens, complicates the task of recovering the state of the dataloader when resuming training. For now, there is NO way to recover the state, but it is possible to achieve this using solutions like [StatefulDataLoader](https://github.com/pytorch/data/tree/main/torchdata/stateful_dataloader) that checkpoint the state of the DataLoader and Dataset. In our case, we would only need to checkpoint the number of samples we have extracted from the Dataset + the token buffers, this for each DP group.
- The most concerning issue today is that to recover the state, we have to use the `.skip()` method. This method is not optimal as it consumes all the samples from the Dataset, but it seems they are [working on a better solution](https://github.com/huggingface/datasets/issues/5380). This is a problem if you have to skip many samples from a XXXL dataset.
- It is very complicated (not impossible) to work with `num_workers > 1`, as the dataset would need to be divided at the worker level, and then it would not be trivial to recover the state if this value changes.

Of all these inconveniences, the one that worries me the most is the third one, but I trust that they will develop an optimal solution soon. We can easily develop solutions for the first and second issues, and the fourth one does not seem too problematic, although we could also address it.
### How Samples Are Produced
[In short](https://github.com/TJ-Solergibert/nanotron/blob/36b1c1814bf71a5fe31f95472001ee0c2ae05ad6/src/nanotron/data/chat_dataset.py#L83), we extract samples from the dataset and apply the chat template until we can no longer fit a full Question-Answer pair into the sequence length of the sample we are constructing. We save this last Question-Answer pair for the next sample and pad the sample we are constructing (In the case of the Llama3 tokenizer as we don't have a pad token we use the <|eot_id|> token). We do this so that each sample has several completed Question-Answer pairs. This packing is greedy, although there are more complex strategies to minimize the number of pad tokens.

The important thing here is that we have developed the [`ChatTokenizer`](https://github.com/TJ-Solergibert/nanotron/blob/36b1c1814bf71a5fe31f95472001ee0c2ae05ad6/src/nanotron/data/chat_tokenizer.py#L6) class to apply the chat template manually and not use solutions like the `apply_chat_template` method of tokenizers. We do this to know at the token level if each one belongs to the assistant's responses or not for the feature of training only on the assistant's tokens. I have added an assert to verify that the result of applying the chat template is exactly the same as the `apply_chat_template` method of tokenizers.
### Dataset Samples
I have developed [this notebook](https://colab.research.google.com/drive/1jrDIc52crI9od9u4Z4s90mEUz5nsuF8g?usp=sharing) so you can check the batches produced by the DataLoader. In summary, the most relevant features are:
> [!NOTE] 
> The `label id` token '-' is actually -100. We switch it because `tokenizer.convert_ids_to_tokens` can't convert '-100' token.
- When training just on the assistant's answers, the first token we predict is **ĊĊ**, which corresponds to "\n\n" from the Llama3 chat template. When interacting with a model we prompt a question + apply chat template and the model starts generating from this "\n\n" token.
Screenshot 2024-05-28 at 02 21 50

- Every Question-Answer pair has its own `position_ids`. This will be relevant later for specifying FA2 to not attend to other samples.
Screenshot 2024-05-28 at 02 22 09

### Other Considerations
- We flatten the MBS to reduce the padding tokens.
- For now, we only support the Open-Orca/SlimOrca dataset format. Check the Keys & Values of the dicts.
- The Open-Orca/SlimOrca is a single turn conversation dataset, so we are appending the <|end_of_text|> token after each assistant's message.
## Collator
The [collator](https://github.com/TJ-Solergibert/nanotron/blob/36b1c1814bf71a5fe31f95472001ee0c2ae05ad6/src/nanotron/data/utils.py#L63) is very similar to `DataCollatorForCLM`, except we now add the `position_ids`. I have also removed several assertions.
## DataLoader
The [DataLoader](https://github.com/TJ-Solergibert/nanotron/blob/36b1c1814bf71a5fe31f95472001ee0c2ae05ad6/src/nanotron/data/dataloader_builder.py#L68) is pretty simple. As I mentioned, it is not trivial to work with `num_workers`.
## Config
I have added a new configuration called [`ChatDatasetsArgs`](https://github.com/TJ-Solergibert/nanotron/blob/36b1c1814bf71a5fe31f95472001ee0c2ae05ad6/src/nanotron/config/config.py#L111) which includes:
- `hf_dataset`: The name or path of the dataset we will stream.
- `hf_dataset_split` ("train"): The split of the dataset.
- `conversation_column_name`: The name of the column from which to extract the conversations.
For debugging purposes, deleted in the final release:
- `train_on_completions_only`: Whether to just train on completions or not.
- `remove_cross_attention`: Whether to just attend to the tokens from the same sample or to all (Vanilla mechanism).

As I already mentioned, the final two configurations will be for evaluating the effect of these two functionalities. I would remove them for the final release since I do not see the benefit of not activating them.

### What Is Still Missing:
- I still need to implement everything related to FA2 and the mechanism to eliminate cross-attention contamination. The most relevant part which belongs to point 2 (No cross-attention contamination between packed samples) can be found [here](https://github.com/OpenAccess-AI-Collective/axolotl/blob/main/src/axolotl/monkeypatch/llama_attn_hijack_flash.py). I hope to finish it this week and present results with comparisons starting next week.

### TODOs:
- Checkpoint DataLoader states to resume training (StatefulDataLoader).
- Support interleaving datasets.
- Improve helper functions.
- Add tests
- Clarify chat dataset format (Dictionary keys)

### DIFF
--- a/examples/chat_config.yaml
+++ b/examples/chat_config.yaml
@@ -1,0 +1,97 @@
+checkpoints:
+  checkpoint_interval: 1000
+  checkpoints_path: /mloscratch/homes/solergib/converter/nanotron/checkpoints
+  checkpoints_path_is_shared_file_system: false
+  resume_checkpoint_path: null
+  save_initial_state: false
+data_stages:
+- data:
+    dataset:
+      hf_dataset: Open-Orca/SlimOrca
+      hf_dataset_split: train
+      conversation_column_name: conversations
+      train_on_completions_only: true
+      remove_cross_attention: true
+    num_loading_workers: 1
+    seed: 42
+  name: General purpose training (Single dataset)
+  start_training_step: 1
+general:
+  benchmark_csv_path: null
+  consumed_train_samples: null
+  ignore_sanity_checks: true
+  project: Chat
+  run: Llama3-8B
+  seed: 42
+  step: null
+lighteval: null
+logging:
+  iteration_step_info_interval: 1
+  log_level: info
+  log_level_replica: info
+model:
+  ddp_bucket_cap_mb: 25
+  dtype: bfloat16
+  init_method:
+    std: 0.025
+  make_vocab_size_divisible_by: 1
+  model_config:
+    bos_token_id: 128000
+    eos_token_id: 128001
+    hidden_act: silu
+    hidden_size: 4096
+    initializer_range: 0.02
+    intermediate_size: 14336
+    is_llama_config: true
+    max_position_embeddings: 8192
+    num_attention_heads: 32
+    num_hidden_layers: 4
+    num_key_value_heads: 8
+    pad_token_id: null
+    pretraining_tp: 1
+    rms_norm_eps: 1.0e-05
+    rope_scaling: null
+    rope_theta: 500000.0
+    tie_word_embeddings: false
+    use_cache: true
+    vocab_size: 128256
+optimizer:
+  accumulate_grad_in_fp32: true
+  clip_grad: 1.0
+  learning_rate_scheduler:
+    learning_rate: 0.0003
+    lr_decay_starting_step: null
+    lr_decay_steps: 98
+    lr_decay_style: cosine
+    lr_warmup_steps: 2
+    lr_warmup_style: linear
+    min_decay_lr: 1.0e-05
+  optimizer_factory:
+    adam_beta1: 0.9
+    adam_beta2: 0.95
+    adam_eps: 1.0e-08
+    name: adamW
+    torch_adam_is_fused: true
+  weight_decay: 0.01
+  zero_stage: 0
+parallelism:
+  dp: 1
+  expert_parallel_size: 1
+  pp: 1
+  pp_engine: 1f1b
+  tp: 1
+  tp_linear_async_communication: false
+  tp_mode: ALL_REDUCE
+profiler: null
+tokenizer:
+  tokenizer_max_length: null
+  tokenizer_name_or_path: meta-llama/Meta-Llama-3-8B-Instruct
+  tokenizer_revision: null
+tokens:
+  batch_accumulation_per_replica: 1
+  limit_test_batches: 0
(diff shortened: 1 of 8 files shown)


xrsrke:
Close. In favor of https://github.com/swiss-ai/nanotron/pull/14

PR #188, opened 2024-05-29
add rope_theta to hf conversion script

jquesnelle:
`rope_theta` was missing from the hf -> nt mapping in the llama conversion script


NouamaneTazi (approved):
LGTM thanks!

PR #202, opened 2024-06-25
refacto generate + use simpler rotary for inference

3outeille:


xrsrke (changes requested):
.

src/nanotron/generation/decode.py
```diff
@@ -772,6 +772,67 @@ def generator():
...
+
+    pipeline_state = PipelineEvalBatchState()
+    with attach_pipeline_state_to_model(model=model, pipeline_state=pipeline_state):
+        batch_size = batch.new_input_ids.shape[0] if use_cache else batch.input_ids.shape[0]
+        seq_len = batch.new_input_ids.shape[1] if use_cache else batch.input_ids.shape[1]
+
+        # Preallocate memory for output logits.
+        logits = None
+        if parallel_context.is_pipeline_last_stage:
+            logits = torch.empty((seq_len, batch_size, model.config.vocab_size), dtype=torch.float32, device=device)
```
xrsrke, line 791: Why do we need to define it here?

PR #207, opened 2024-07-08
Add layer-wise activation recomputation to llama model

C-TC:
When benchmarking Nanotron with big models (e.g. Llama2-70b) we found that Nanotron consumes more memory than Megatron under the same parallel configuration. Besides the fix https://github.com/huggingface/nanotron/pull/203, we also add layer-wise activation recomputation (also known as gradient checkpointing) to mitigate this issue. Currently it is controlled by flag `recompute_layer` in `src/nanotron/config/parallelism_config.py`. 

Megatron also offers more fine-grained control over activation recomputation, but I found it is not straightforward to apply that to Nanotron mainly because of the pipelineblock abstraction. It would be nice if developers could shed some light on this. And I would be glad to help formalize this PR.


3outeille:
lgtm !

PR #208, opened 2024-07-18
Memory optimization in async tp-linear

AleHD:
This PR introduces the memory optimization methods implemented in #203, but now allows them to be used in the async comm regime. This PR also includes commits related to fixing the row-parallel tp-linear in #172, so both of those PRs should be merged first to make reviewing this one easier. This PR also includes two modes: recomputing or not the all_gather. Recomputing is more-memory efficient, but slightly slower. 

Here is a table that summarizes my observations on a tp4 llama8b model on A100 GPUs:
| Method | Average throughput (tok/sec/gpu) | Max memory reserved (GB) |
| :- | :-: | :-: |
| Baseline (current main implementation) | 3415 | 72.6 |
| Sync no-recompute | 3525 (+3%) | 68.5 (-6%) |
| Sync recompute | 3427 | 61.9 (-15%) |
| Async no-recompute | 3587 (+5%) | 68.8 (-5%) |
| Async recompute | 3526 (+3%) | 60.9 (-16%) |

These changes should prove to be very useful for more efficient training. I recommend using the async-recompute setting, but using async-norecompute might make more sense for the extra throughput when memory is not a concern. In addition, as dp and pp increase and optimizer states and parameters become more sharded, the memory savings this PR brings should only increase as these affect memory activations. Very useful for scaling to larger models.

I attach the wandb logs for llama8b (top) experiments and a tiny 152M model (bottom) to study the effects on smaller models.

image
image
(**blue** = sync baseline, **green** = sync no-recompute, **purple** = sync recompute, **yellow** = async baseline, **gray** = async recompute, **red** = async no-recompute).


src/nanotron/parallel/tensor_parallel/functional.py
```diff
@@ -141,22 +142,27 @@ def forward(ctx, tensor, weight, bias, group, tp_mode):
                 # `tensor` can sometimes not be contiguous
                 # https://cs.github.com/pytorch/pytorch/blob/2b267fa7f28e18ca6ea1de4201d2541a40411457/torch/distributed/nn/functional.py#L317
                 tensor = tensor.contiguous()
-                ctx.save_for_backward(tensor, weight)
+                # ctx.save_for_backward(tensor, weight)
```
3outeille, line 145: remove comments

src/nanotron/config/config.py
```diff
@@ -111,7 +111,7 @@ def __post_init__(self):
 class DataArgs:
     """Arguments related to the data and data files processing"""
 
-    dataset: Union[PretrainDatasetsArgs, NanosetDatasetsArgs]
+    dataset: Optional[Union[PretrainDatasetsArgs, NanosetDatasetsArgs]]
```
xrsrke, line 114: why is "dataset` optional?

xrsrke (approved):
LGTM!

3outeille:
lgtm as well

PR #220, opened 2024-08-22
add support for s3 checkpoint

eliebak:
# Goal
Allow the upload checkpoint to s3 during training + resuming from ckpt

# How
Add `S3UploadArgs` class, define like this in the config 
```
s3_upload:
  remove_after_upload: true
  s5cmd_concurrency: 5
  s5cmd_numworkers: 16
  s5cmd_path: path_to_s5cmd 
  upload_s3_path: path_to_s3
```
Add support for s3 path in `resume_checkpoint_path` and it will copy (with s5cmd) it too`checkpoints_path`. This is done in the `pre_init` phase with the `parse_ckpt_path(config=self.config, parallel_context=self.parallel_context)` function.

### DIFF
--- a/src/nanotron/utils.py
+++ b/src/nanotron/utils.py
@@ -3,6 +3,7 @@ import inspect
 import os
 import random
 import socket
+import re
 from contextlib import ExitStack, contextmanager
 from typing import ContextManager, List, Optional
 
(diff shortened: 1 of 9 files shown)


src/nanotron/utils.py
```diff
@@ -3,6 +3,7 @@
 import os
 import random
 import socket
+import re
```
3outeille, line 6: remove

3outeille:
Please run the pre-commit so that it fix automatically everything related to coding styles 

```
pip install pre-commit
pre-commit install
```

PR #238, opened 2024-11-02
Load random states from checkpoint

gritukan:
Model's random states are saved into checkpoint but not loaded during recovery which cause non-deterministic behavior.


NouamaneTazi:
We're actually thinking about removing random states as they're not very pytorch native.
Instead we'd recommend using REDUCE_SCATTER (which is sequence parallelism) that ensure determinism instead

PR #246, opened 2024-11-21
Optimize memory when loading checkpoint

NouamaneTazi:
Offload memory states to CPU to avoid memory peak when loading checkpoint

### DIFF
--- a/src/nanotron/optim/base.py
+++ b/src/nanotron/optim/base.py
@@ -1,5 +1,5 @@
 from abc import ABC, abstractmethod
-from typing import Any, Callable, Dict, List, Optional, Set, TypeVar
+from typing import Any, Callable, Dict, List, Optional, Set, TypeVar, Union
 
 import torch
 
@@ -34,7 +34,7 @@ class BaseOptimizer(ABC):
         ...
 
     @abstractmethod
-    def load_state_dict(self, state_dict: dict) -> None:
+    def load_state_dict(self, state_dict: dict, map_location: Optional[Union[str, torch.device]] = None) -> None:
         ...
 
     @abstractmethod
--- a/src/nanotron/sanity_checks.py
+++ b/src/nanotron/sanity_checks.py
@@ -263,7 +263,7 @@ def check_optim_state_in_sync(optimizer: optim.BaseOptimizer, pg: dist.ProcessGr
     for _, optim_state in sorted(optimizer.state_dict()["state"].items(), key=lambda x: x[0]):
         for name, tensor in optim_state.items():
             if name == "step":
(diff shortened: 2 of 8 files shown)


xrsrke (approved):
LGTM. Left a small comment.

src/nanotron/optim/base.py
```diff
@@ -46,3 +67,112 @@ def inherit_from(self, cls) -> bool:
...
+    else:
+        if param.is_floating_point():
+            return value.to(dtype=param.dtype, device=target_device)
+        else:
+            return value.to(device=target_device)
+
+
+# Modified from torch.optim.Optimizer.load_state_dict
+@torch._disable_dynamo
+def custom_load_state_dict(self, state_dict: StateDict, map_location: Union[str, torch.device]) -> None:
```
xrsrke, line 108: It would be nice to avoid duplicating code from PyTorch, in case they introduce internal changes in future versions. Could we create a context manager that automatically moves tensors within a given function to a target device?

```python
@contextmanager
def ensure_cpu():
    original_device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
    try:
        torch.set_default_device('cpu')  # This currently requires manual tracking
        yield
    finally:
        torch.set_default_device(original_device)

# Usage
with ensure_cpu():
    tensor = torch.tensor([1.0, 2.0, 3.0])
    print(tensor.device)  # Outputs: cpu

```

src/nanotron/sanity_checks.py
```diff
@@ -259,12 +263,11 @@ def after_optim_step_sanity_checks(
...
 
-def check_optim_state_in_sync(optimizer: optim.BaseOptimizer, pg: dist.ProcessGroup):
-    for _, optim_state in sorted(optimizer.state_dict()["state"].items(), key=lambda x: x[0]):
+def check_optim_state_in_sync(optim_state_dict: dict, pg: dist.ProcessGroup):
+    for _, optim_state in sorted(optim_state_dict["state"].items(), key=lambda x: x[0]):
         for name, tensor in optim_state.items():
             if name == "step":
-                tensor = tensor.to("cuda")
-
+                continue
```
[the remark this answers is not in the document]
NouamaneTazi: I think we don't care if step in cuda or cpu (because if when optim states is created on GPU, torch creates step on CPU)

src/nanotron/optim/base.py
NouamaneTazi: As discussed, `torch.set_default_device` won't help us create the optimizer states in gpu (optim states are created specifically on the model's device)
I tried a bunch of options before this and none seemed to work. This ensures that we create optim states directly on the device we want

Issue #237, opened 2024-10-31
Nanoset reuses same shuffling for repeated data (epochs)


Lauler:
Nanoset's index builder does not re-shuffle dataset and sample indices within epochs when training secondary, third, etc epochs. It instead concatenates a copy of the same indices for any repeated data. See:

https://github.com/huggingface/nanotron/blob/51ca40bc5e1b1f5dcb55eaeb0b6f86dda03f3979/src/nanotron/data/nanoset.py#L114-L124

I would suggest changing this so the shuffling is applied within each epoch before concatenating the data. Here's the edited version I use:

```python
# Shuffle the indices within each epoch and concatenate them
r = np.random.RandomState(self.random_seed)
epoch_random_seeds = r.randint(0, 2**32 - 1, num_epochs)
dataset_indices = []
dataset_sample_indices = []
for i in range(num_epochs):
    # Shuffle the sample and dataset indices in epoch with same seed
    numpy_random_state = np.random.RandomState(epoch_random_seeds[i])
    numpy_random_state.shuffle(dataset_index)
    numpy_random_state = np.random.RandomState(epoch_random_seeds[i])
    numpy_random_state.shuffle(dataset_sample_index)

    dataset_indices.append(dataset_index)
    dataset_sample_indices.append(dataset_sample_index)

# Concatenate the within-epoch shuffled indexes
dataset_index = np.concatenate(dataset_indices)
dataset_sample_index = np.concatenate(dataset_sample_indices)
```

If you think this is reasonable I can submit a pull request.

TJ-Solergibert:
Yes! I [warned](https://github.com/huggingface/nanotron/blob/51ca40bc5e1b1f5dcb55eaeb0b6f86dda03f3979/docs/nanoset.md?plain=1#L110) this in the little docs I wrote a while back. 

Looks good to me, make sure that the unit tests are passing and try to submit a PR!

PR #247, opened 2024-11-24
Add shuffling in Nanotron for subsequent epochs when data is repeated

Lauler:
Nanoset's index builder does not re-shuffle dataset and sample indices within epochs when training secondary, third, etc epochs. It instead concatenates a copy of the same indices for any repeated data. This PR adds unique within-epoch shuffling for each epoch.

See the following issue: https://github.com/huggingface/nanotron/issues/237 

I ran the tests in `tests/nanotron`:

```
=========================================================================== warnings summary ===========================================================================
tests/helpers/context.py:7: 35 warnings
  /home/faton/projects/text/nanotron_dev/nanotron/tests/helpers/context.py:7: PytestCollectionWarning: cannot collect test class 'TestContext' because it has a __init__ constructor (from: nanoset/test_build_nanoset_dataloader.py)
    class TestContext:

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=================================================================== 12 passed, 35 warnings in 44.55s ===================================================================
```

### DIFF
--- a/src/nanotron/data/nanoset.py
+++ b/src/nanotron/data/nanoset.py
@@ -111,14 +111,26 @@ class Nanoset(torch.utils.data.Dataset):
         dataset_index, dataset_sample_index = build_nanoset_index_helper(
             n_samples=samples_per_epoch, weights=self.dataset_weights, dataset_sizes=self.dataset_lengths
         )
-        # Shuffle the indexes the same way
-        numpy_random_state = np.random.RandomState(self.random_seed)
-        numpy_random_state.shuffle(dataset_index)
-        numpy_random_state = np.random.RandomState(self.random_seed)
-        numpy_random_state.shuffle(dataset_sample_index)
-        # Concatenate num_epochs the shuffled indexes
-        dataset_index = np.concatenate([dataset_index for _ in range(num_epochs)])
-        dataset_sample_index = np.concatenate([dataset_sample_index for _ in range(num_epochs)])
+
+        # Shuffle indices in each epoch with different random seeds and concatenate them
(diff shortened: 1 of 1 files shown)


src/nanotron/data/nanoset.py
```diff
@@ -111,14 +111,26 @@ def build_nanoset_index(self) -> np.ndarray:
...
-        numpy_random_state.shuffle(dataset_index)
-        numpy_random_state = np.random.RandomState(self.random_seed)
-        numpy_random_state.shuffle(dataset_sample_index)
-        # Concatenate num_epochs the shuffled indexes
-        dataset_index = np.concatenate([dataset_index for _ in range(num_epochs)])
-        dataset_sample_index = np.concatenate([dataset_sample_index for _ in range(num_epochs)])
+
+        # Shuffle indices in each epoch with different random seeds and concatenate them
+        r = np.random.RandomState(self.random_seed)
+        epoch_random_seeds = r.randint(0, 2**32 - 1, num_epochs)
```
NouamaneTazi, line 117: Can we just use `self.random_seed + num_epoch` for easier reproducibility?

NouamaneTazi (approved):
Nice catch! LGTM 🤗

src/nanotron/data/nanoset.py
Lauler: Good suggestion, makes it a lot simpler and clearer what's going on. I've changed the code to incorporate this in latest commit:

https://github.com/huggingface/nanotron/blob/f060414194e02c2e22e8fa0ffd8326363917a796/src/nanotron/data/nanoset.py#L116-L123

PR #253, opened 2024-11-28
resuming checkpoint without lr schedule or optimizer state

eliebak:
quick pr to add the possibility to resume training without the optimizer state or lr schedule for continual learning

### DIFF
--- a/src/nanotron/trainer.py
+++ b/src/nanotron/trainer.py
@@ -190,7 +190,7 @@ class DistributedTrainer:
             optimizer_args=self.config.optimizer,
             parallel_context=self.parallel_context,
         )
-        if self.init_checkpoint_path is not None:
+        if self.init_checkpoint_path is not None and self.config.checkpoints.load_optimizer:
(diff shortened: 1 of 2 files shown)


src/nanotron/trainer.py
```diff
@@ -215,7 +215,7 @@ def __init__(
             )
 
         # Define iteration start state
-        if self.init_checkpoint_path is not None:
+        if self.init_checkpoint_path is not None and self.config.checkpoints.load_lr_scheduler:
```
NouamaneTazi, line 218: is this only lr related?

NouamaneTazi (commented):
Nice! left a qst

src/nanotron/trainer.py
eliebak: I think when you want to do a new lr schedule, you also want to do a different data stages + reset the "last_train_step" argument

NouamaneTazi (approved):
LGTM!

PR #243, opened 2024-11-17
Fix initial_lr when resuming training

Lauler:
The argument initial_lr in lambda_lr() is initialized incorrectly from the **learning rate** of an optimizer's parameter groups. This causes the learning rate to be set incorrectly when models are resumed from checkpoints trained with standard parametrization LR schedulers.

It should instead be initialized from the **initial learning rate** of the optimizer's param groups.

See this issue comment for context: https://github.com/huggingface/nanotron/issues/233#issuecomment-2481663592

PR #256, opened 2024-11-29
Fix wrong initialization of lr scheduler

kylematoba:
The initialization of the learning rate scheduler does not correctly resume. This is because the optimizer state is loaded first, and the lambda in the LR scheduler is partialied to the LR this implies. In this PR I fix(?) it by first initializing both the optimizer and the LR scheduler, then load the state dict for each. 

So, for example, when you save the state dict at a step you have 
```
lr_scheduler.lr_lambdas[0]
functools.partial(<function lr_scheduler_builder.<locals>.lr_lambda at 0x7f6e1c539fc0>, initial_lr=0.0003)
```

but when you reload it to resume you have 
```
lr_scheduler.lr_lambdas[0]
functools.partial(<function lr_scheduler_builder.<locals>.lr_lambda at 0x7f00c053b760>, initial_lr=1e-05)
```

Obviously, using a totally wrong LR schedule makes resumption useless. 

It seems to work, except for the 15 step example `examples/config_tiny_llama.yaml`, where it's close to working, but not exact. I'm assuming this is due to some sort of warm up or something. If it's not clear to you I can look into it further. 

Plus a small fix for the code just not running in `serialize/main.py`.

### DIFF
--- a/src/nanotron/serialize/main.py
+++ b/src/nanotron/serialize/main.py
@@ -60,6 +60,7 @@ def save(
     should_save_lr_scheduler: bool = True,
     sanity_checks: bool = True,
 ) -> None:
+
     assert isinstance(training_metadata, TrainingMetadata)
 
     try:
@@ -107,6 +108,7 @@ def save(
                 lr_scheduler=lr_scheduler,
                 parallel_context=parallel_context,
                 root_folder=root_folder,
+                is_zero=config.optimizer.zero_stage
             )
     except Exception as e:
         log_rank(
--- a/src/nanotron/trainer.py
+++ b/src/nanotron/trainer.py
@@ -190,6 +190,13 @@ class DistributedTrainer:
             optimizer_args=self.config.optimizer,
             parallel_context=self.parallel_context,
         )
+        
+        # Init learning rate scheduler
+        self.lr_scheduler = lr_scheduler_builder(
+            optimizer=self.optimizer,
+            lr_scheduler_args=self.config.optimizer.learning_rate_scheduler,
+            total_training_steps=self.config.tokens.train_steps,
+        )
         if self.init_checkpoint_path is not None:
             load_optimizer(
                 optimizer=self.optimizer,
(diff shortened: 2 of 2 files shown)


NouamaneTazi (changes requested):
LGTM! please rebase main and remove `lr_scheduler._initial_step()` as shown here https://github.com/huggingface/nanotron/pull/243#issuecomment-2517482125

kylematoba:
@NouamaneTazi that's done.

std-song:
I reproduced this on the current `main` branch at `2411b022` (Python 3.10.8, PyTorch 2.1.2).

The resume path still advances the scheduler by one step:

```
checkpoint: last_epoch=7, lr=9.572050015e-4
uninterrupted next update uses: 9.572050015e-4
current resume before next update: last_epoch=8, lr=9.055061048e-4
parameter error after one resumed update: 5.1677e-05
```

The root cause matches the earlier discussion:

1. The optimizer state is loaded before `lr_scheduler_builder`, so the LR lambda captures the checkpoint's current LR instead of the configured base LR.
2. `load_lr_scheduler` calls the private `_initial_step()`, which increments `last_epoch` and changes the optimizer LR before the next optimizer update.

I prepared a clean fix against current `main` that:

- builds the scheduler before loading optimizer state;
- loads scheduler state without advancing it;
- restores optimizer param-group LRs from `scheduler.get_last_lr()`;
- adds CPU regression tests comparing uninterrupted vs resumed LR, `last_epoch`, and the next parameter update.

Verification:

- the regression test fails on unmodified `main` with `assert 8 == 7`;
- 2 new regression tests pass after the fix;
- existing runnable serialization tests: 7 passed, 3 skipped (one remaining test hardcodes TP=2 and the validation host has one GPU);
- Ruff passes.

Since this PR is still open but has merge conflicts and is based on an older tree, would you prefer that I open a fresh, minimal PR against current `main`, or contribute the tests/fix through this PR?

PR #260, opened 2024-12-05
Updating dependency versions in pyproject.toml and README.

sapiosaturn:
Hi,

I was trying to get the example in the README to run, and encountered a few issues. They were eventually resolved through dependency version changes, so I thought I'd make a PR.

**README change — specifying <python 3.12** 
This was motivated by the following issue that I faced when using Python 3.12.
python importerror caused by version incompatibility

**pyproject.toml changes — specified flash-attn<2.70**
In the upcoming flash-attn 2.7.0, the return format for `bert_padding.unpad_input` was changed, which causes the example to fail with flash-attn 2.7.0.

Screenshot of commit changing return format from flash-attn repo:
commit message from flash-attn repository showing the change

**pyproject.toml changes — specified numpy<2**
NumPy 2 had some breaking changes, and one of those changes seems to break the serializing in nanotron. This is illustrated in the screenshot below, where a numpy int seems to have been saved as `np.int64(48)` instead of 48. Specifying numpy<2 solves this issue. 

Screenshot of issue:
illustration of issue caused by numpy 2

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Nanotron is a library for pretraining transformer models. It provides a simple a
 ## Installation
 
 ```bash
-# Requirements: Python>=3.10
+# Requirements: Python>=3.10,<3.12
 git clone https://github.com/huggingface/nanotron
 cd nanotron
 pip install --upgrade pip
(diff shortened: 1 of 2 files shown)


NouamaneTazi (approved):
Perfect! Ty for the PR

PR #278, opened 2025-02-05
Add MLA 

zzhhjjj:
Add MLA to Nanotron

### DIFF
--- a/src/nanotron/models/llama.py
+++ b/src/nanotron/models/llama.py
@@ -17,6 +17,7 @@
 from typing import Dict, List, Optional, Union
 
 import torch
+import torch.nn.functional as F
 from torch import nn
 from torch.utils.checkpoint import CheckpointFunction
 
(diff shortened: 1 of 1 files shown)


zzhhjjj:
# Config example

```yaml
model:   
  ddp_bucket_cap_mb: 25
  dtype: bfloat16
  init_method:
    std: 0.025
  make_vocab_size_divisible_by: 1  
  model_config:  
    ...
    vocab_size: 50272
    # MLA
    q_lora_rank: 128 
    kv_lora_rank: 128
    qk_nope_head_dim: 128
    qk_rope_head_dim: 128
    v_head_dim: 128
```

examples/mla/mla_dp=8.yaml
```diff
@@ -0,0 +1,107 @@
...
+  dp: 8
+  expert_parallel_size: 1
+  pp: 1
+  pp_engine: 1f1b
+  recompute_layer: false
+  tp: 1
+  tp_linear_async_communication: false
+  tp_mode: ALL_REDUCE
+  tp_recompute_allgather: false
+  # domino:
```
xrsrke, line 93: remove :)

src/nanotron/models/llama.py
```diff
@@ -701,8 +849,9 @@ def __init__(
         layer_idx: int,
     ):
         super().__init__()
+        Attention = MLA if config.kv_lora_rank is not None else CausalSelfAttention
```
xrsrke, line 852: name it attn_cls?

examples/mla/mla_dp=8.yaml
```diff
@@ -0,0 +1,107 @@
+checkpoints:
+  checkpoint_interval: 100000
+  checkpoints_path: /fsx/haojun/long_context_weights/test
+  checkpoints_path_is_shared_file_system: false
+  resume_checkpoint_path: null
+  load_lr_scheduler: false
+  load_optimizer: false
+  save_final_state: true
+  save_initial_state: false
+data_stages:
```
xrsrke, line 10: rename the file to something like llama_with_mla maybe

src/nanotron/config/config.py
```diff
@@ -228,6 +228,10 @@ def __post_init__(self):
             self.dtype = cast_str_to_torch_dtype(self.dtype)
 
         self.model_config._is_using_mup = isinstance(self.init_method, SpectralMupInit)
+        if self.model_config.kv_lora_rank is not None:
```
xrsrke, line 231: maybe assert either kv_lora_rank or num_key_value_heads isn't None

src/nanotron/models/llama.py
```diff
@@ -692,6 +692,154 @@ def forward(
...
 
+class MLA(nn.Module):
+    def __init__(
+        self,
+        config: LlamaConfig,
+        parallel_config: Optional[ParallelismArgs],
+        tp_pg: dist.ProcessGroup,
+        layer_idx: int,
+    ):
+        super().__init__()
```
xrsrke, line 703:
```suggestion
    def __init__(
        self,
        config: LlamaConfig,
        parallel_config: Optional[ParallelismArgs],
        tp_pg: dist.ProcessGroup,
        layer_idx: int,
    ):
        """
        Implementation of DeepSeek's MLA
        Section 2.1.1. Multi-Head Latent Attention
        DeepSeek-V3 Technical Report
        https://arxiv.org/abs/2412.19437
        """
        super().__init__()
```

src/nanotron/models/llama.py
```diff
@@ -692,6 +692,154 @@ def forward(
...
+        kv = self.kv_up(self.kv_norm(kv))  # [seq_len, batch_size, n_local_heads * (qk_nope_head_dim + v_head_dim)]
+        kv = kv.view(seq_len, batch_size, self.n_local_heads, self.qk_nope_head_dim + self.v_head_dim)
+        k_nope, v = torch.split(
+            kv, [self.qk_nope_head_dim, self.v_head_dim], dim=-1
+        )  # [seq_len, batch_size, n_local_heads, qk_nope_head_dim], [seq_len, batch_size, n_local_heads, v_head_dim]
+        k = torch.cat(
+            [k_nope, k_pe.expand(-1, -1, self.n_local_heads, -1)], dim=-1
+        )  # [seq_len, batch_size, n_heads, qk_head_dim]
+
+        q = q.transpose(0, 1).contiguous()  # [batch_size, seq_len, n_heads, qk_head_dim]
```
xrsrke, line 824: would be nice if we could do these shape operations using einops, because it would be easier to read https://github.com/arogozhnikov/einops

xrsrke (changes requested):
Overall, it looks good, but I recommend adding unit tests to sanity-check MLA with different values of tp_mode, async_communication, and tp_recompute_allgather, and make sure that the output shape of the MLA class is as expected

src/nanotron/config/config.py
zzhhjjj: I think num_key_value_heads has a default value, so it's not None

src/nanotron/models/llama.py
zzhhjjj: In Transformers, they don't do that. It's better to follow this practice?

xrsrke (approved):
LFG

src/nanotron/models/llama.py
```diff
@@ -692,6 +692,160 @@ def forward(
...
+            self.qk_head_dim == self.v_head_dim
+        ), "MLA only supports equal query/key and value head dimensions for now"
+
+        # Initialize rotary embedding
+        self.rotary_embedding = RotaryEmbedding(
+            dim=self.qk_rope_head_dim, end=config.max_position_embeddings, theta=config.rope_theta
+        )
+
+        # Initialize linear layers
+        self.q_down = nn.Linear(self.dim, self.q_lora_rank, bias=False)
```
NouamaneTazi, line 754: Add a small warning here please saying that defining q_down as a nn.Linear means it'll be duplicated on all GPUs then we'll run our `mark_unsharded_params_as_tied_across_tp` function https://github.com/zzhhjjj/nanotron/blob/014589e826850e76a2341724f3bd17279ab48adb/src/nanotron/trainer.py#L984 to mark these linears tied, so that we sync their grads in backward
P.S: need to double check this is how MLA works

src/nanotron/models/llama.py
```diff
@@ -692,6 +692,160 @@ def forward(
...
+    ):
+        seq_len, batch_size, _ = hidden_states.shape
+
+        q = self.q_up(self.q_norm(self.q_down(hidden_states)))
+        q = q.view(seq_len, batch_size, self.n_local_heads, self.qk_head_dim)
+        q_nope, q_pe = torch.split(
+            q, [self.qk_nope_head_dim, self.qk_rope_head_dim], dim=-1
+        )  # [seq_len, batch_size, n_local_heads, qk_nope_head_dim], [seq_len, batch_size, n_local_heads, qk_rope_head_dim]
+        q_pe = (
+            self.rotary_embedding(q_pe.transpose(0, 1), position_ids=None).transpose(0, 1).contiguous()
```
NouamaneTazi, line 808: is the transpose(0,1) needed here?

src/nanotron/models/llama.py
```diff
@@ -692,6 +692,160 @@ def forward(
...
+        q = torch.cat(
+            [q_nope, q_pe.expand(-1, -1, self.n_local_heads, -1)], dim=-1
+        )  # [seq_len, batch_size, n_heads, qk_head_dim]
+
+        kv = self.kv_down(hidden_states)  # [seq_len, batch_size, qk_rope_head_dim + kv_lora_rank]
+        kv, k_pe = torch.split(
+            kv, [self.kv_lora_rank, self.qk_rope_head_dim], dim=-1
+        )  # [seq_len, batch_size, kv_lora_rank], [seq_len, batch_size, qk_rope_head_dim]
+        k_pe = (
+            self.rotary_embedding(k_pe.unsqueeze(2).transpose(0, 1), position_ids=None).transpose(0, 1).contiguous()
```
NouamaneTazi, line 819: ditto

src/nanotron/models/llama.py
```diff
@@ -701,8 +855,9 @@ def __init__(
         layer_idx: int,
     ):
         super().__init__()
+        attn_cls = MLA if config.kv_lora_rank is not None else CausalSelfAttention
```
NouamaneTazi, line 858: I'd rather make it more explicit like use `config.use_mla` here and assert somewhere that the other configs (e.g. kv_lora_rank) are well defined. This can be done in `config.py`

src/nanotron/parallel/tensor_parallel/functional.py
```diff
@@ -414,7 +414,7 @@ def backward(ctx, grad_output: torch.Tensor):
         grad_weight = grad_output.T @ total_input
         grad_input = grad_output @ weight
         if group.size() == 1:
-            sub_grad_input = grad_input
+            sub_grad_input = grad_input.reshape(input_size)  # [s*b, h_in] -> [s, b, h_in]
```
NouamaneTazi, line 417: interesting! why do we need this? what kind of error we got?

NouamaneTazi (changes requested):

NouamaneTazi (changes requested):
Left some comments!! Looking nice alrdy

src/nanotron/models/llama.py
zzhhjjj: yes. otherwise the results would be different

src/nanotron/models/llama.py
zzhhjjj: it seems a little bit redundant to me, because kv_lora_rank = MLA in this case

src/nanotron/parallel/tensor_parallel/functional.py
zzhhjjj: shape mismatch. It breaks backward pass

src/nanotron/models/llama.py
zzhhjjj: checked it, and it's tied

src/nanotron/models/llama.py
zzhhjjj: Yes, transpose is slow. This is the first time I'm using q.transpose in MLA, so I'm not sure if I can avoid a transpose here

zzhhjjj: I see what you mean. Now it's doing less transpose

src/nanotron/config/models_config.py
```diff
@@ -55,6 +55,14 @@ class LlamaConfig:
     use_cache: bool = True
     vocab_size: int = 32000
 
+    #  MLA
+    # https://huggingface.co/deepseek-ai/DeepSeek-V3-Base/blob/main/config.json
+    q_lora_rank: Optional[int] = None
+    kv_lora_rank: Optional[int] = None
+    qk_nope_head_dim: Optional[int] = None
+    qk_rope_head_dim: Optional[int] = None
+    v_head_dim: Optional[int] = None
```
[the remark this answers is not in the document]
zzhhjjj: I'm following this: https://huggingface.co/deepseek-ai/DeepSeek-V3/blob/main/config.json

PR #284, opened 2025-02-24
[Feature] DoMiNO with 62% communication hiding in tensor parallelism

xrsrke:
We are overlapping 75% in the forward pass and 50% in the backward pass (previously 75% for the backward pass), so overall 62% overlapping

- reference run: https://wandb.ai/neuralink/nanotron_domino/runs/csjwnvka?nw=nwuserneuralink
- domino run: https://wandb.ai/neuralink/nanotron_domino/runs/rpcoqaf6

image

the first PR is ready for review (i split it to two PRs), some left work for the next PR

- intra-layer overlapping (current overlapping communication within a layer), but if we do intra-layer overlapping, then we can almost totally overlapping the comm
- create an fixed buffer to concat hidden states (1st image)
- double check if there is cuda stream switching's overhead (4.3.1 in the 2nd image)
- minimize kernel launch overhead (cuda graph? 4.3.2 in 2nd image)

profiling results

image

image

`/fsx/phuc/new_workspace/experiments/nanotron_domino/profilings/exp0bc4_domino_profiling_for_exp2a1c7c2b/20250221-183933/ip-26-0-161-103_455950.1740163192231207296.pt.trace.json`


xrsrke:
Move here: https://github.com/huggingface/nanotron/pull/285

PR #286, opened 2025-03-01
[Feature] Over 99% communication overlap in Tensor Parallelism using Domino

hwchen2017:
Hi nanotron team,  this is the follow-up work of #285. This PR implements the cross layer communication computation overlapping,  and it can achieve over 99% communication overlapping on single node. Using Domino, TP training on single node is almost communication "free". 

### Profiling results

Forward pass
  
image

Backward pass
image

Please note that the flag enabling Domino is hardcoded [here](https://github.com/hwchen2017/nanotron/blob/hongwei_domino_cross_layer_overlap/src/nanotron/models/llama.py#L52). It needs to be configurable for ease of use, but I haven't done it yet. There might also be other considerations on your side. You can further improve the code quality based on this branch. Please let me know if you have any other questions.

Cc: @GuanhuaWang

### DIFF
--- a/examples/domino_tiny_llama.yaml
+++ b/examples/domino_tiny_llama.yaml
@@ -1,0 +1,98 @@
+checkpoints:
+  checkpoint_interval: 2000
+  checkpoints_path: checkpoints
+  checkpoints_path_is_shared_file_system: false
+  resume_checkpoint_path: null
+  save_initial_state: false
+data_stages:
+- data:
+    dataset:
+      dataset_overwrite_cache: false
+      dataset_processing_num_proc_per_process: 1
+      hf_dataset_config_name: null
+      hf_dataset_or_datasets: stas/openwebtext-10k
+      hf_dataset_splits: train
+      text_column_name: text
+    num_loading_workers: 1
+    seed: 42
+  name: Stable Training Stage
+  start_training_step: 1
+general:
+  benchmark_csv_path: null
+  consumed_train_samples: null
+  ignore_sanity_checks: true
+  project: debug
+  run: tiny_llama_%date_%jobid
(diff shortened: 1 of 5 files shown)


xrsrke:
Hi, thanks for the PR. I did a bit of profiling and saw that the forward pass is impressive, as the communication fully overlaps, but the backward pass isn't as good

image

image

xrsrke:
Thanks for the PR! I'm working on integrating your inter-layer into this PR https://github.com/huggingface/nanotron/pull/285, where we overlap the backward pass better and have also tested the convergence

GuanhuaWang:
@xrsrke great catch! if possible, let's do it step-by-step (by merging this cross-layer overlap forward with some not so good backward, lets call it ver0.1.0). 

We have noted down this issue and will provide solutions later (maybe later call the fixed as ver 0.1.1 then). Feel free to discuss any plan here, we are more than happy to help! 

> Hi, thanks for the PR. I did a bit of profiling and saw that the forward pass is impressive, as the communication fully overlaps, but the backward pass isn't as good
> 
> image
> 
> image

hwchen2017:
Hi @xrsrke , the performance of backward pass can be further optimized. But optimization of BWD pass will have conflicts with FWD pass (We will explain it later in out updated paper), and it makes the optimal e2e optimization complicated.

We may need to have different implementations for different models/config/nodes. And it's not finalized yet. You may still see some bubbles in computation stream for both FWD and BWD pass depending on the settings. 

This branch can bring pretty good improvement over async megatron, and it doesn't have the potential data race problem. It's a good start point to think about the integration with your existing code. We will let you know once a better solution is available.

PR #328, opened 2025-03-31
Nope

eliebak:
# What does this PR do?
Option do not do rope on some layers (see https://arxiv.org/abs/2501.1879)


src/nanotron/config/models_config.py
```diff
@@ -141,6 +141,7 @@ class Qwen2Config:
     sliding_window_size: Optional[int] = None
     z_loss_enabled: bool = False  # Z-loss regularization https://www.jmlr.org/papers/volume24/22-1144/22-1144.pdf
     z_loss_coefficient: float = 0.0001  # Default from the paper (10^-4)
+    no_rope_layer: Optional[int] = None
```
NouamaneTazi, line 144: maybe a small comment explaining that we'll remove rope from multiple of it?

NouamaneTazi (approved):

NouamaneTazi (approved):
Clean ✨

PR #335, opened 2025-04-02
Nanoset clean up old index

eliebak:
# What does this PR do?
Fallback to the old indexing for now, add the eos, add the return position args in nanoset args 
To use with: https://github.com/huggingface/datatrove/tree/dataloader_positions

### DIFF
--- a/run_train.py
+++ b/run_train.py
@@ -188,7 +188,7 @@ def get_dataloader_from_data_stage(
             tokenizer_path = trainer.config.tokenizer.tokenizer_name_or_path
             tokenizer = AutoTokenizer.from_pretrained(tokenizer_path)
             eos_token_id = tokenizer.eos_token_id
-            assert eos_token_id is not None and data.dataset.return_positions is True, "Tokenizer must have a eos token if return_positions is True"
+            assert eos_token_id is not None and data.dataset.return_positions is True, "Tokenizer must have an eos token if return_positions is True"
             log_rank(
                 f"[Nanoset] Creating Nanoset with {len(data.dataset.dataset_folder)} dataset folders and {trainer.config.tokens.train_steps * trainer.global_batch_size} train samples",
                 logger=logger,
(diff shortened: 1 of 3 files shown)


run_train.py
```diff
@@ -196,10 +199,15 @@ def get_dataloader_from_data_stage(
                 train_split_num_samples=trainer.config.tokens.train_steps * trainer.global_batch_size,
                 dataset_weights=data.dataset.dataset_weights,
                 random_seed=data.seed,
+                return_positions=data.dataset.return_positions,
+                eos_token_id=trainer.model_config.eos_token_id,
```
NouamaneTazi, line 203: can we use `trainer.config.tokenizer.tokenizer_name_or_path` here instead?

NouamaneTazi (changes requested):

run_train.py
eliebak: we need to load the tokenizer if we do that, so i guess we can use model_config.eos_token_id and do a sanity check in the post_init of the config ? i will add it in another pr if that's ok

eliebak: will clean the model config from eos/bos and moove it to the tokenizerArgs directly in another PR i agree it make more sense

NouamaneTazi: yes i think it should be done in run_train.py similar to the HF dataset case [here](https://github.com/huggingface/nanotron/blob/25b4376fa59cade0ba47c0aba49e807e283577da/run_train.py#L118)
`

NouamaneTazi: just load tokenizer and pass tokenizer.eos_token_id (with an assert that ` tokenizer.eos_token_id is not None`)

NouamaneTazi (changes requested):

run_train.py
eliebak: fixed

src/nanotron/data/nanoset.py
```diff
@@ -48,6 +50,9 @@ def __init__(
         # Init
         self.dataset_folders = dataset_folders
         self.sequence_length = sequence_length
+        self.eos_token_id = eos_token_id
+        self.return_positions = return_positions
```
NouamaneTazi, line 54: we can add a small comment here before merging explaining that return_positions here means we get position_ids from nanoset/datatrove

NouamaneTazi (approved):
LGTM! lets add a small comment before merging.. thx a lot

PR #363, opened 2025-05-08
fix makefile, sync with datatrove, update lighteval config

hynky1999:
## Fixes
- Allows uv users to compile dataset cpp lib (previously it would fail as by default it fall backs to non-uv python as uv doesn't have python3-config)
- Changes few defaults in lighteval config + removes the token setting which doesn't make any sense
- Fixes types of ModelConfig to get errors if union can't match
- Updates the tokenized dataset logic to reflect changes in datatrove (caching is now handled by dattrove)


NouamaneTazi (approved):
LGTM! thx

